Issue #9, opened 2019-03-11
Feature Request: RSA Blind signing


phayes:
Hi there,

I was wondering if you would be interested in implementing RSA blind-signing in this crate?  

Blind-signing is rare enough that it won't be picked up by any of the mainstream cryptography crates (like ring, openssl etc) so it might make sense to include it here.

dignifiedquire:
It already exists :) https://github.com/RustCrypto/RSA/blob/master/src/key.rs#L416

phayes:
Hi there, 

Thanks for the quick reply.   What I mean specifically, is to allow the following workflow:

1. Alice blinds the hash of a message, generating a blinded message and an unblinder.
2. Alice sends the blind message to the signer Bob. Bob the signer does not know what the message is, since it is blind.  But Bob can still sign the blind message.
3. Alice then "unblinds" the signature (using the unblinder) so that the signature now signs the original message, and not the blind-message that was sent to the signer. 

This type of thing is super useful in instances where someone needs to sign a message without knowing the contents of the message.  This is useful in, for example, online voting where we want to sign the vote as coming from a valid voter, but keep the contents of the vote secret. 

I've implemented such a scheme in Go (https://github.com/cryptoballot/rsablind), and would love to have the same things in Rust.

dignifiedquire:
Very interesting, I did not know about this scheme. In general I would say this is something I would be happy to add, but I can't promise that I will have time to implement it soon . 

Is there an RFC or a paper detailing the scheme that can be used as a reference?

phayes:
Hi @dignifiedquire,

I've done a bunch more research into this, and there isn't really a "standard" for doing RSA blind-signing.   RSA-FDH is pretty close, but it has never been formalized into a spec, so I don't think it belongs in this crate.

I'm going to be creating a new crate called "RSA-FDH" that implements a blind-signing scheme, but it will be a scheme specific to that crate, and not part of any standard. 

What we CAN do in this crate however, is break the core blinding and unblinding functions out to make building an RSA blind-signing crate easier.  I've done that in this PR: #10

phayes:
Closing this. 

RSA blind-signing is implemented in the RSA-FDH crate: https://github.com/phayes/rsa-fdh

PR #10, opened 2019-03-14
Splitting blind and unblind into their own functions, exposing raw RSA functions via `internals` module

phayes:
I've split blinding and unblinding into their own public functions.   

This relates to #9 - instead of implementing a full blind-signing protocol in this crate, we just expose the `blind` and `unblind` functions to make building such a protocol easier. 

In addition I've re-exported the raw RSA functions into a new module called 'danger'.   The intention here is to allow other crates that are creating custom RSA schemes to access raw RSA functions, but to ward off casual users.

### DIFF
--- a/src/danger.rs
+++ b/src/danger.rs
@@ -1,0 +1,6 @@
+pub use crate::key::blind;
+pub use crate::key::check_public;
+pub use crate::key::decrypt;
+pub use crate::key::decrypt_and_check;
+pub use crate::key::encrypt;
+pub use crate::key::unblind;
--- a/src/danger.rs
+++ b/src/danger.rs
@@ -1,6 +1,0 @@
-pub use crate::key::blind;
-pub use crate::key::check_public;
-pub use crate::key::decrypt;
-pub use crate::key::decrypt_and_check;
-pub use crate::key::encrypt;
-pub use crate::key::unblind;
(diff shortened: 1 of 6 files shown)


phayes:
I just found an error in this. don't merge yet!  Fix is coming up.

src/lib.rs
```diff
@@ -51,9 +51,22 @@ extern crate serde_test;
...
 extern crate sha1;
 
+/// Useful algorithms.
 pub mod algorithms;
+
+/// Contains raw RSA functions that should not be used directly.
+///
+/// This module is provided so that other crates that implement non-standard RSA schemes can use these RSA primitives.
+/// If you are not implementing a custom RSA-based scheme, you should not use these functions.
+pub mod danger;
```
dignifiedquire: what do you think of putting this export behind a compile feature?

src/danger.rs
```diff
@@ -0,0 +1,6 @@
+pub use crate::key::blind;
```
dignifiedquire: I think it makes sense to actually pull these methods into this module and call it something like internal and import them in key.rs

phayes:
Hi @dignifiedquire , 

I've followed your suggestion and made a new module called "internals", and optionally made it public behind a feature flag called "expose-internals". 

The only thing I'm a bit unsure on is how to handle precomputed values, and if they should stay in the `key` module or be moved to the `internals` module.  For now I oped to keep them in the `key` module and mark them `pub(crate)`.

dignifiedquire:
Thank you

PR #11, opened 2019-03-17
adds the to_public_key() method for RSAPrivateKey

phayes:
While generally passing around RSAPrivateKey as a `PublicKey` trait is optimal, there are some instances where you want to deal in concrete types and not be passing around the secret private data.  This PR adds a utility `extract_public()` method for RSAPrivateKey for those circumstances.

### DIFF
--- a/src/key.rs
+++ b/src/key.rs
@@ -286,7 +286,7 @@ impl RSAPrivateKey {
     ///
     /// Generally this is not needed since `RSAPrivateKey` implements the `PublicKey` trait,
     /// but it can occationally be useful to discard the private information entirely.
-    pub fn to_public(&self) -> RSAPublicKey {
+    pub fn to_public_key(&self) -> RSAPublicKey {
         // Safe to unwrap since n and e are already verified.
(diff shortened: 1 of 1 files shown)


src/key.rs
```diff
@@ -282,6 +282,15 @@ impl RSAPrivateKey {
         k
     }
 
+    /// Extract the public key from the private key, cloning `n` and `e`.
+    ///
+    /// Generally this is not needed since `RSAPrivateKey` implements the `PublicKey` trait,
+    /// but it can occationally be useful to discard the private information entirely.
+    pub fn extract_public(&self) -> RSAPublicKey {
```
dignifiedquire: I would prefer `to_public_key` to keep with rust naming conventions

dignifiedquire:
Thanks, also note that if you don't need the private key anymore, you can do the following already

```rust
let public_key: RSAPublicKey = my_private_key.into();
```

src/key.rs
phayes: Done!

phayes:
Now called `to_public_key`.   :)

PR #14, opened 2019-03-18
Moving to 2018 edition

phayes:
This pull request does two things:

1. Moves to the 2018 edition and applies 2018 edition idioms
2. Runs rustfmt

The application of rustfmt makes this PR look scarier than it is, so you may want to checkout 558d34f, review the diff at 558d34f, and then and run rustfmt yourself.


dignifiedquire:
Thank you

PR #22, opened 2019-08-18
feat: nostd, core+alloc support

roblabla:
Blocked on https://github.com/dignifiedquire/num-bigint/pull/12

Missing: docs, tests and a way to avoid having base64 break our builds to rust-lang/cargo#4866.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -41,8 +41,9 @@ hex = "0.4.0"
 serde_test = "1.0.89"
 rand_xorshift = "0.2.0"
 pem = "0.8"
-#sha-1 = "0.9.0"
-#sha3 = "0.9.0"
+sha-1 = { default-features = false, version = "0.9.0" }
+sha2 = { default-features = false, version = "0.9.0" }
+sha3 = { default-features = false, version = "0.9.0" }
 
 [[bench]]
 name = "key"
@@ -59,5 +60,5 @@ nightly = ["subtle/nightly", "num-bigint/nightly"]
 serde = ["num-bigint/serde", "serde_crate"]
 serde1 = ["serde"] # deprecated
 expose-internals = []
-std = ["alloc", "simple_asn1", "digest/std"]
+std = ["alloc", "simple_asn1", "digest/std", "rand/std"]
 alloc = ["digest/alloc"]
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -12,19 +12,17 @@ categories = ["cryptography"]
 readme = "README.md"
 
 [dependencies]
-num-bigint = { version = "0.6", features = ["rand", "i128", "u64_digit", "prime", "zeroize"], package = "num-bigint-dig" }
-num-traits = "0.2.6"
-num-integer = "0.1.39"
-num-iter = "0.1.37"
-lazy_static = "1.3.0"
-rand = "0.7.0"
-byteorder = "1.3.1"
-thiserror = "1.0.11"
-subtle = "2.0.0"
-simple_asn1 = "0.4"
+num-bigint = { version = "0.6", features = ["i128", "u64_digit", "prime", "zeroize"], default-features = false, package = "num-bigint-dig" }
+num-traits = { version= "0.2.9", default-features = false, features = ["libm"] }
+num-integer = { version = "0.1.39", default-features = false }
+num-iter = { version = "0.1.37", default-features = false }
+lazy_static = { version = "1.3.0", features = ["spin_no_std"] }
+rand = { version = "0.7.0", default-features = false }
+byteorder = { version = "1.3.1", default-features = false }
+subtle = { version = "2.0.0", default-features = false }
+simple_asn1 = { version = "0.4", optional = true }
 pem = { version = "0.8", optional = true }
-digest = { version = "0.9.0", features = ["std"] }
-sha2 = "0.9.0"
+digest = { version = "0.9.0", default-features = false }
 
 [dependencies.zeroize]
 version = "1.1.0"
@@ -35,7 +33,7 @@ package = "serde"
 optional = true
 version = "1.0.89"
 default-features = false
-features = ["std", "derive"]
+features = ["derive"]
 
 [dev-dependencies]
 base64 = "0.12.0"
@@ -43,8 +41,8 @@ hex = "0.4.0"
 serde_test = "1.0.89"
 rand_xorshift = "0.2.0"
 pem = "0.8"
-sha-1 = "0.9.0"
-sha3 = "0.9.0"
+#sha-1 = "0.9.0"
+#sha3 = "0.9.0"
 
 [[bench]]
 name = "key"
@@ -56,8 +54,10 @@ name = "key"
 # debug = true
 
 [features]
-default = ["pem"]
+default = ["std", "pem"]
 nightly = ["subtle/nightly", "num-bigint/nightly"]
 serde = ["num-bigint/serde", "serde_crate"]
 serde1 = ["serde"] # deprecated
 expose-internals = []
+std = ["alloc", "simple_asn1", "digest/std"]
+alloc = ["digest/alloc"]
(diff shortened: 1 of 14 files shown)


Cargo.toml
```diff
@@ -50,7 +51,8 @@ name = "key"
 # debug = true
 
 [features]
-default = []
+default = ["std"]
+std = []
```
dignifiedquire: shouldn’t this enable the std features on the crates that got it disabled above?

newpavlov:
I wonder if we can write a custom library for stack allocating bigints bounded by a prime constant. it would be also useful for SRP implementation. Do you know any prior art libraries which have done it?

dignifiedquire:
doesnt smallvec already do this, given it is the backing structure of the bigints?

roblabla:
That's exactly what smallvec does. A `SmallVec<[u8; 4096]>` will store up to 4096 bytes on the stack, and then switch to heap-allocated array.

newpavlov:
I meant bigints without the heap fallback. Since most operations are done modulo some value with a known length, it should be doable? Plus this hypothetical crate can be a significantly smaller, since we don't need general purpose bigint.

roblabla:
Well for statically allocated vecs, [arrayvec](https://docs.rs/arrayvec/0.4.11/arrayvec/) exists and is pretty good. But it would require changing the bigint library to be backend agnostic somehow, since otherwise we'd have to copy paste a fairly significant portion of the bigint algorithms over.

dignifiedquire:
@roblabla I have updated `num-bigint-dig`, want to update this PR so we can ship it?

roblabla:
I'll try to squeeze some time during the week to get this up and running.

roblabla:
I did some fixes here and there, but I'm stuck against a wall again. `base64` is enabling the std feature of `byteorder`, which breaks no_std build. Now, base64 is a dev dependency, but because cargo is, frustratingly, still susceptible to [unifying dev-dependencies and dependencies features](https://github.com/rust-lang/cargo/issues/4664). 

A PR was merged in August [allowing nostd usage of base64](https://github.com/marshallpierce/rust-base64/pull/111), but there hasn't been a new release of base64 since. So we're waiting for upstream to [release a new version](https://github.com/marshallpierce/rust-base64/issues/118).

tarcieri:
@roblabla you can use `subtle_encoding::base64` instead of `base64`, which is fully `no_std` friendly, has no dependencies on crates with activated `std` features, and provides constant-time encoding and decoding if you do end up using it for private keys: https://github.com/iqlusioninc/crates/tree/develop/subtle-encoding

roblabla:
The new base64 dependency breaks Rust 1.31 builds ;_;. Subtle-encoding also has an MSRV of 1.36. May I bump the MSRV @dignifiedquire ?

dignifiedquire:
there is always sth :D feel free to bump the version

src/algorithms.rs
```diff
@@ -2,6 +2,7 @@ use num_bigint::traits::ModInverse;
 use num_bigint::{BigUint, RandPrime};
 use num_traits::{FromPrimitive, One, Zero};
 use rand::Rng;
+use libm::F64Ext;
```
dignifiedquire: Could you update to libm@0.2, I did that in num-bigint-dig, unfortunately that doesn't have the extension anymore, so you need to do some manual wrapping of things.

.travis.yml
```diff
@@ -2,14 +2,16 @@ language: rust
 
 matrix:
   include:
-    - rust: 1.31.0
+    - rust: 1.34.0
```
dignifiedquire: I bumped the msr to 1.36, because of smallvec@1.0 in #33

dignifiedquire:
@roblabla merged everything else on master, any chance you want to pick this up again and bring it over the finishline?

roblabla:
I'll take a look!

roblabla:
Some quick thoughts:

- RSA now relies on thiserror. This crate is not no-std compatible. I'll probably have to rip it out and implement the traits by hand.
- Regarding the possibility of a future no-std no-alloc version, the RSA crate now makes some fairly heavy use of the `DynDigest` trait, which relies on liballoc. In theory it could be swapped with a trait that takes a buffer instead of returning a boxed slice, and panics if the buffer is the wrong size.
- The parse functionality will have to be gated on libstd for now, since simple_asn1 is not no_std-compatible. I don't think people working in no_std environments are likely to need PEM parsing though, so that shouldn't be a big issue.

tarcieri:
You can also use [`displaydoc`](https://crates.io/crates/displaydoc) if you'd like a proc macro for impl'ing `Display` for the error type which supports `no_std`.

roblabla:
I ended up using cargo expand and copy pasting the output of thiserror lol. I'll look into displaydoc. I pushed a first draft of the no-std support. It's WIP and untested, but at least it builds.

Things to do:

- I added an "alloc" feature, the idea being that down the line, we might want a no-std no-alloc build to be possible. In order to support this use-case without introducing a breaking change, I figured the best idea would be to add an alloc feature *now*, and force the build to fail if either std or alloc feature aren't given. Does that sound good?
- I'll try to get qemu setup to run the tests on ARM on CI.

Things that will happen in a follow-up PR:
- ASN1 parsing in core+alloc environment.
- Find a way to ditch `lazy_static`. That dependency is a pain in the neck, it uses a feature to enable libcore support (instead of the convention of adding a feature to add libstd support), which means we force lazy_static to spinlock even in std environments.

tarcieri:
+1 to `no_std` + `alloc`. Many of the other RustCrypto crates support features which use this combination.

roblabla:
Ooh, I hadn't heard of conquer-once. And it does seem to provide lazy init on no_std: https://docs.rs/conquer-once/0.2.0/conquer_once/spin/type.Lazy.html is available in no-std, and is exactly what we want. We should be able to swap between this and the top-level `Lazy` depending on whether we enabled the std feature!

tarcieri:
Is this ready to merge? Looks good to me at first glance

.travis.yml
```diff
@@ -11,5 +11,7 @@ script:
   - if [[ "$TRAVIS_NIGHTLY" == "true" ]]; then cargo test --verbose --all-features; fi
   - if [[ "$TRAVIS_NIGHTLY" == "true" ]]; then cargo bench --verbose --all-features --no-run; fi
   - if [[ "$TRAVIS_NIGHTLY" != "true" ]]; then cargo test --verbose; fi
+  - if [[ "$TRAVIS_NIGHTLY" != "true" ]]; then rustup target add thumbv7m-none-eabi; fi
+  - if [[ "$TRAVIS_NIGHTLY" != "true" ]]; then cargo build --verbose --no-default-features --features=alloc --target thumbv7m-none-eabi; fi
```
dignifiedquire, line 15: Did you locally test if this passes tests, while building is great, it is pretty easy to have things failing still. Ideally we would use `cross` to actually run the tests 
Something like

```
cross test --no-default-features --features alloc --target thumbv7m-none-eabi
```

should work if we can get cross to work on travis

tarcieri: I might suggest trying to switch to GitHub Actions first before you do this.

Namely [`actions-rs` supports relatively painless cross-compilation](https://github.com/actions-rs/cargo#cross-compilation) with `use_cross: true`.

roblabla: I agree moving to github actions seems ideal, but it should be done in a separate PR IMO. I'll try to open a gh action migration PR this week end.

dignifiedquire: github actions pr is merged

roblabla: Right so this is fun. Cross can't run tests for thumbv7m-none-eabi, since that target doesn't have libstd or libtest. I could build and test for the arm-linux-gnu target, but we still need to have a build step for thumbv7-none-eabi to avoid having accidental dependencies on libstd.

roblabla: So I changed it up by building for thumb, and testing on arm.

dignifiedquire:
@roblabla almost there one more rebase please and can you check out those clippy warnings please? Sorry for the long turnaround times

roblabla:
Should be good to go I hope!

PR #25, opened 2019-08-21
Reexport biguint

roblabla:
BigUint is part of the public API surface of RSA. As such, I think it's a good idea to reexport it. It avoids forcing the user to add a dependency on num-bigint-dig when importing pubkeys or privkeys.


dignifiedquire:
thanks, makes sense

PR #26, opened 2019-08-21
Implement PSS

roblabla:
I took a stab at implementing PSS. It's based on the Go implementation. All the internals are done and should be working. It's built on top of the no_std PR but I can decouple them if necessary.

While implementing it, I found quite a few problems with the current interface of the RSA crate that makes implementing PSS in a way that doesn't break compatibility very painful.

The first problem is that PSS is parametrized by the hash algorithm (and actually hashes things, unlike PKCS1 v1.5), but verify and sign are generic over the Hash trait, where Hash does not actually allow us to hash things (it simply gives the hash size and the algorithmID). Ideally, verify would work like:

```rust
fn verify<T: Digest + AlgorithmId>(
        &self,
        padding: PaddingScheme,
        hashed: &[u8],
        sig: &[u8])
```

This loses the ability to sign/verify "raw" unhashed data for PKCS1 v1.5, but maybe it could be put in a separate function? Furthermore, it requires implementing an AlgorithmId trait for all the supported hashes for PKCS1 v1.5. (Maybe Digest could have an `algorithm_oid()` function? That's kind of meh though).

Another problem is that PSS signing really needs an RNG. Ideally, sign would take the RNG as a parameter. I guess we could use `thread_rng()` internally, but that makes the function unusable in no_std context.

Yet another problem is that it's impossible to pass the salt length to the internal algorithm from the verify/sign functions. I currently default to the "AUTO" mode from the go library, but that makes the implementation less generic. I guess I could modify the PaddingScheme enum to pass implementation-specific arguments (such as the salt len). If we go down that road, the AlgorithmID for PKCS1 v1.5 could also go there.

TBH, I wonder if it would be worth it to just have different functions for the different padding schemes. They all have different requirements and parameters, and would allow us to give the "best" interface for each of them.

------------

While we're talking about breaking changes, I think we'll need to change the public API anyways if integration with RustCrypto/signatures is a goal. See RustCrypto/signatures#25 .


Cargo.toml
```diff
@@ -22,6 +22,9 @@ byteorder = { version = "1.3.1", default-features = false }
 failure = { version = "0.1.5", default-features = false, features = ["derive"] }
 subtle = { version = "2.0.0", default-features = false }
 libm = "0.1.4"
+signature = { version = "0.2.0", default-features = false }
+sha2 = { version = "0.8", default-features = false }
+sha-1 = { version = "0.8.1", default-features = false }
```
dignifiedquire: General note, I would really like to not have hard dependencies on hashing crates. Not sure if it can be avoided.

roblabla:
So I picked this up again. I decided to play around and see what design I would end up with if I completely ignored backwards compatibility and just redesigned the interface. This is where I arrived at. This currently manages to avoid depending on the hashing crates.

dignifiedquire:
I have integrated this into the existing api in https://github.com/RustCrypto/RSA/pull/43

roblabla:
awesome! I'll check it out.

dignifiedquire:
closing in favor of #43

Issue #34, opened 2020-01-02
New API design


str4d:
Prompted by https://github.com/RustCrypto/signatures/issues/25#issuecomment-525248346 and my desire to get #18 and #26 merged :smile:

The main design difference compared to the current traits is that instead of making the "padding scheme" a parameter of the pubkey sign/verify functions, we make each scheme a first-class primitive that wraps the common public or private key. It needs to be an explicit choice by the user anyway.

Current draft proposal (as of 2020-03-07) (without changes to the `signature` crate, and without handling the encryption cases): 
```rust
// This module has been merged into master
mod raw {
    pub trait EncryptionPrimitive {
        /// Do NOT use directly! Only for implementors.
        fn raw_encryption_primitive(&self, plaintext: &[u8]) -> Result<Vec<u8>>;
    }

    pub trait DecryptionPrimitive {
        /// Do NOT use directly! Only for implementors.
        fn raw_decryption_primitive<R: Rng>(
            &self,
            rng: Option<&mut R>,
            ciphertext: &[u8],
        ) -> Result<Vec<u8>>;
    }
}

mod key {
    pub trait PublicKeyParts {
        /// Returns the modulus of the key.
        fn n(&self) -> &BigUint;
        /// Returns the public exponent of the key.
        fn e(&self) -> &BigUint;
        /// Returns the modulus size in bytes. Raw signatures and ciphertexts for
        /// or by this public key will have the same size.
        fn size(&self) -> usize {
            (self.n().bits() + 7) / 8
        }
    }

    pub trait PrivateKey: crate::raw::DecryptionPrimitive + PublicKeyParts {
        /// Could have functions like this for usability?
        pub fn sign_pkcs1v15(&self) -> crate::pkcs1v15::Signer;
        pub fn sign_pkcs1v15_blinded<R: Rng>(&self, rng: R) -> crate::pkcs1v15::Signer;
        pub fn sign_pss(&self) -> crate::pss::Signer;
        pub fn sign_pss_blinded(&self) -> crate::pss::Signer;
    }

    pub trait PublicKey: crate::raw::EncryptionPrimitive + PublicKeyParts {
        /// Could have functions like this for usability?
        pub fn verify_pkcs1v15(&self) -> crate::pkcs1v15::Verifier;
        pub fn verify_pss(&self) -> crate::pss::Verifier;
    }

    pub struct RSAPrivateKey { ... }

    impl crate::raw::DecryptionPrimitive for RSAPrivateKey { ... }

[...]

Issue #32, opened 2019-12-04
Swappable RSADP for YubiKey support


str4d:
While working on [`yubikey-piv`](https://github.com/iqlusioninc/yubikey-piv.rs) I was testing the RSA logic by implementing OAEP around it. What I ended up doing was copying in all the logic from #18, and replacing:
```rust
    let mut em = {
        let mut c = BigUint::from_bytes_be(ciphertext);
        let mut m = internals::decrypt(rng, priv_key, &c)?;
        let em = internals::left_pad(&m.to_bytes_be(), k);

        c.zeroize();
        m.zeroize();

        em
    };
```
with
```rust
    let mut em = {
        let m = yubikey.decrypt_data(&ciphertext, algorithm, slot).unwrap();
        // I forgot to check if the padding was actually necessary
        left_pad(&m, k)
    };
```

It seems like the way to achieve this more generally (across all decryption schemes) would be with a trait of the form:
```rust
trait PrivateKey {
    /// Do NOT use directly! Only for implementors.
    fn raw_decryption_primitive<R: Rng>(
        &self,
        rng: Option<&mut R>,
        ciphertext: &[u8],
    ) -> Result<Vec<u8>>;

    /// Decrypt the given message.
    fn decrypt(&self, padding: PaddingScheme, ciphertext: &[u8]) -> Result<Vec<u8>> {
        ...
    }

    /// Decrypt the given message.
    /// Uses `rng` to blind the decryption process.
    pub fn decrypt_blinded<R: Rng>(
        &self,
        rng: &mut R,
        padding: PaddingScheme,
        ciphertext: &[u8],
    ) -> Result<Vec<u8>> {
        ...
    }

    fn decrypt_oaep(...) -> Result<...> {
        oaep::decrypt(...)
    }
}

impl PrivateKey for RSAPrivateKey {
    fn raw_decryption_primitive<R: Rng>(
        &self,
        rng: Option<&mut R>,
        ciphertext: &[u8],
    ) -> Result<Vec<u8>> {
        let mut c = BigUint::from_bytes_be(ciphertext);
        let mut m = internals::decrypt(rng, priv_key, &c)?;
        let em = internals::left_pad(&m.to_bytes_be(), k);

        c.zeroize();
        m.zeroize();

        em
    }
}
```

Then in `yubikey-piv` we could do something like:
```rust
struct YubiKeyRsaPrivateKey {
    yubikey: &mut YubiKey,
    algorithm: AlgorithmId,
[...]

PR #35, opened 2020-01-04
Initial key trait refactor

str4d:
This is part of #32, and a precursor for #34.

The new trait structure:
- `PublicKeyParts` - provides access to the modulus, exponent, and size.
- `EncryptionPrimitive` - raw RSA encryption, for implementers.
- `DecryptionPrimitive` - raw RSA decryption, for implementers.
- `PublicKey: EncryptionPrimitive + PublicKeyParts` - the public key API for users. Home to `encrypt` and `verify` functions.
- `PrivateKey: DecryptionPrimitive + PublicKeyParts` - the private key API for users. Will eventually be home to `decrypt` and `sign` functions.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -62,6 +62,7 @@ pub mod padding;
 
 mod key;
 mod pkcs1v15;
+mod raw;
 
 pub use self::key::{PublicKey, RSAPrivateKey, RSAPublicKey};
 pub use self::padding::PaddingScheme;
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -8,20 +8,21 @@
 //! extern crate rsa;
 //! extern crate rand;
 //!
-//! use rsa::{PublicKey, RSAPrivateKey, PaddingScheme};
+//! use rsa::{PublicKey, RSAPrivateKey, RSAPublicKey, PaddingScheme};
 //! use rand::rngs::OsRng;
 //!
 //! let mut rng = OsRng;
 //! let bits = 2048;
-//! let key = RSAPrivateKey::new(&mut rng, bits).expect("failed to generate a key");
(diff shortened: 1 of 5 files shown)


str4d:
Note that `PrivateKey` here is an abstraction over a private key. In the case of a hardware token, the concrete implementation would hold the modulus, exponent, and e.g. an active connection to the hardware token. This is a different abstraction compared to e.g. having a `SigningBackend` trait as suggested in https://github.com/RustCrypto/RSA/issues/34#issuecomment-570389758.

str4d:
A subsequent refactor would either modify `PublicKey::verify` to take a typed `Signature`, or split it into concrete functions like `PublicKey::{verify_pkcs1v15, verify_pss}` (and similarly for `PrivateKey::sign`). These are the two directions I mention in https://github.com/RustCrypto/RSA/issues/34#issuecomment-570413461.

src/lib.rs
```diff
@@ -8,20 +8,21 @@
...
 //!
-//! use rsa::{PublicKey, RSAPrivateKey, PaddingScheme};
+//! use rsa::{PublicKey, RSAPrivateKey, RSAPublicKey, PaddingScheme};
 //! use rand::rngs::OsRng;
 //!
 //! let mut rng = OsRng;
 //! let bits = 2048;
-//! let key = RSAPrivateKey::new(&mut rng, bits).expect("failed to generate a key");
+//! let private_key = RSAPrivateKey::new(&mut rng, bits).expect("failed to generate a key");
+//! let public_key = RSAPublicKey::from(private_key.clone());
```
dignifiedquire: This should probably work without calling `clone`, as this means duplicating a bunch of secrets, when we actually only want the public parts

dignifiedquire:
Merged in #42

Issue #30, opened 2019-11-14
Panic at 'invalid coeff' using a private key generated by openssl


CBenoit:
Hi,

I tried playing around with RSA and everything is nice as long as I generate my private keys with the API provided by the crate itself (great!).
[...]

PR #36, opened 2020-01-09
fix: make precompute optional instead of panicking on invalid coeff

bcortier-devolutions:
Currently, `RSAPrivateKey::precompute` panic when Chinese remainder theorem can't be applied.
This is a frequent issue with openssl / mbedtls generated private keys. See issue https://github.com/RustCrypto/RSA/issues/30.
However, it is not required to be called to properly create signatures.

This PR
- modify the `precompute` method to return a `Result` instead of panicking
- ignore the result of the `precompute` call when building a RSAPrivateKey using `from_components`. That is if Chinese remainder theorem can't be applied, there is no precomputed data, but no panic either. The private key can still be used for signing operations.

This isn't a breaking change as long as user code doesn't `#[deny(unused_must_use)]` when calling `precompute`.

### DIFF
--- a/src/errors.rs
+++ b/src/errors.rs
@@ -23,6 +23,8 @@ pub enum Error {
     InvalidModulus,
     #[fail(display = "invalid exponent")]
     InvalidExponent,
+    #[fail(display = "invalid coefficient")]
+    InvalidCoefficient,
     #[fail(display = "public exponent too small")]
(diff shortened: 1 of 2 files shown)


bcortier-devolutions:
@dignifiedquire Okay, here it is!

PR #38, opened 2020-02-13
Add parsing for pkcs1 and pkcs8 encoded rsa keys

### LINKED ISSUE #37: How do I actually import a public key from a string ?
In go you simply `pem.Decode` and then `x509.ParsePKIXPublicKey` the given byte slice and you get a `rsa.PublicKey`. However, in the rust world none of those libraries are actually compatible with each other so how do I import an existing public key?

aloucks:
Adds simple functions for parsing `pkcs1` and `pkcs8` encoded keys and a `pem` feature that re-exports the `pem` crate with `TryFrom` implementations.

The `TryFrom` implementations will auto-detect `pkcs1` or `pkcs8` via the header tags.

```rust
let pem = rsa::pem::parse(private_key_pem)?;
let private_key = RSAPrivateKey::try_from(pem)?;

let pem = rsa::pem::parse(public_key_pem)?;
let public_key= RSAPublicKey::try_from(pem)?;
```
Or, manually:

```rust
// -----BEGIN RSA PRIVATE KEY-----
let private_key = rsa::parse_private_key_pkcs1(pkcs1_private_key_base64_decoded)?;
// -----BEGIN RSA PUBLIC KEY-----
let public_key = rsa::parse_public_key_pkcs1(pkcs1_public_key_base64_decoded)?;

// -----BEGIN PRIVATE KEY-----
let private_key = rsa::parse_private_key_pkcs8(pkcs8_private_key_base64_decoded)?;
// -----BEGIN PUBLIC KEY-----
let public_key = rsa::parse_public_key_pkcs8(pkcs8_public_key_base64_decoded)?;
```

Fixes #37

EDIT:

Bikeshed: Should the `rsa::parse_[public|private]_key_pkcs[1|8] ` functions be prefixed with `decode_` instead?

### DIFF
--- a/src/key.rs
+++ b/src/key.rs
@@ -199,6 +199,30 @@ impl RSAPublicKey {
     /// following a `-----BEGIN RSA PUBLIC KEY-----` header.
     ///
     /// <https://tls.mbed.org/kb/cryptography/asn1-key-structures-in-der-and-pem>
+    /// 
+    /// # Example
+    /// 
+    /// ```
+    /// use rsa::RSAPublicKey;
+    /// 
+    /// # // openssl rsa -pubin -in tiny_key.pub.pem -RSAPublicKey_out
+    /// let file_content = r#"
+    /// -----BEGIN RSA PUBLIC KEY-----
+    /// MEgCQQCuWe45NapeQ6rkb5T5hYMgQwr3T0NE9rItqDwFJjcNmf6m9kq/wRAaFqWo
+    /// BX/BpdSuD0YqSUrnQ5ejf1XW9gmJAgMBAAE=
+    /// -----END RSA PUBLIC KEY-----
+    /// "#;
+    /// 
+    /// let der_encoded = file_content
+    ///     .lines()
+    ///     .filter(|line| !line.starts_with("-"))
+    ///     .fold(String::new(), |mut data, line| {
+    ///         data.push_str(&line);
+    ///         data
+    ///     });
+    /// let der_bytes = base64::decode(&der_encoded).expect("failed to decode base64 content");
+    /// let public_key = RSAPublicKey::from_pkcs1(&der_bytes).expect("failed to parse key");
+    /// ```
     pub fn from_pkcs1(der: &[u8]) -> Result<RSAPublicKey> {
         crate::parse::parse_public_key_pkcs1(der)
     }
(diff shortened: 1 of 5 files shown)


aloucks:
The build is failing on the `hex` crate with rust `1.36`. I believe this is unrelated to the pull request:

```
error: enum variants on type aliases are experimental
  --> /home/travis/.cargo/registry/src/github.com-1ecc6299db9ec823/hex-0.4.1/src/error.rs:30:13
   |
30 |             Self::InvalidStringLength => write!(f, "Invalid string length"),
   |             ^^^^^^^^^^^^^^^^^^^^^^^^^
error: aborting due to 3 previous errors
error: Could not compile `hex`.
```

tarcieri:
Thanks for this, and looks like some unrelated fixes are needed for the `hex` crate.

This implementation looks okay to me. I'm not super wild about the `rsa::parse_[public|private]_key_pkcs` functions.

I think those should probably be methods on the individual key types. See the [`FromPkcs8`](https://docs.rs/signatory/0.18.0/signatory/encoding/pkcs8/trait.FromPkcs8.html) trait in the `signatory` crate as a potential example of an API.

The `TryFrom` impls look good to me.

src/lib.rs
```diff
@@ -60,11 +60,16 @@ pub mod hash;
...
+#[cfg(feature="pem")]
+pub use pem;
+
 mod key;
 mod pkcs1v15;
+mod parse;
 
 pub use self::key::{PublicKey, RSAPrivateKey, RSAPublicKey};
 pub use self::padding::PaddingScheme;
+use self::parse::{parse_private_key_pkcs1, parse_private_key_pkcs8, parse_public_key_pkcs1, parse_public_key_pkcs8};
```
tarcieri: It seems like this is unnecessary?

aloucks: I've removed it and used the full crate path where referenced.

tarcieri:
Cool, well other than the unrelated errors from `hex` this looks good to me.

@dignifiedquire @str4d thoughts?

I think we can probably simplify the implementation and eliminate the `simple_asn1` dependency by vendoring just the ASN.1 DER support we need and looking for a [hardcoded PKCS#8 header](https://github.com/RustCrypto/RSA/issues/37#issuecomment-581475260) ala BearSSL and the corresponding translation in the [`ecdsa` crate](https://github.com/RustCrypto/signatures/blob/master/ecdsa/src/convert.rs#L57), but I think that's a debatable tradeoff and not worth blocking this PR over.

src/key.rs
```diff
@@ -192,6 +192,26 @@ impl RSAPublicKey {
...
         Ok(k)
     }
+
+    /// Parse a `PKCS1` encoded RSA Public Key.
+    ///
+    /// The `der` data is expected to be the `base64` decoded content
+    /// following a `-----BEGIN RSA PUBLIC KEY-----` header.
+    ///
+    /// <https://tls.mbed.org/kb/cryptography/asn1-key-structures-in-der-and-pem>
+    pub fn from_pkcs1(der: &[u8]) -> Result<RSAPublicKey> {
```
dignifiedquire: could you add a doctesz for this?

src/parse.rs
```diff
@@ -0,0 +1,407 @@
...
+#[cfg(feature = "pem")]
+impl TryFrom<pem::Pem> for RSAPrivateKey {
+    type Error = Error;
+
+    /// Parses a `PKCS8` or `PKCS1` encoded RSA Private Key.
+    ///
+    /// Expects one of the following `pem` headers:
+    /// - `-----BEGIN PRIVATE KEY-----`
+    /// - `-----BEGIN RSA PRIVATE KEY-----`
+    fn try_from(pem: pem::Pem) -> Result<RSAPrivateKey> {
```
dignifiedquire: could you add a doctest for this, so it is easy to see in the docs how it works?

src/parse.rs
```diff
@@ -0,0 +1,407 @@
...
+#[cfg(feature = "pem")]
+impl TryFrom<pem::Pem> for RSAPublicKey {
+    type Error = Error;
+
+    /// Parses a `PKCS8` or `PKCS1` encoded RSA Public Key.
+    ///
+    /// Expects one of the following `pem` headers:
+    /// - `-----BEGIN PUBLIC KEY-----`
+    /// - `-----BEGIN RSA PUBLIC KEY-----`
+    fn try_from(pem: pem::Pem) -> Result<RSAPublicKey> {
```
dignifiedquire: same here

aloucks: @dignifiedquire I've added examples for all of the new functions.

aloucks:
@dignifiedquire @tarcieri Anything else needed before merging?

dignifiedquire:
@aloucks sorry for the delay, I want to fix CI before merging, but it looks good thank you 👌

aloucks:
@dignifiedquire A new version of `hex` was published and the build is passing now.

tarcieri:
:shipit:

PR #42, opened 2020-03-06
Key trait refactor

dignifiedquire:
Rebased and fixed the nit I had from #35. 

@str4d @tarcieri I think this is a good first step and am going to merge this, unless you have objections


str4d:
No objections from me.

PR #43, opened 2020-03-06
Integrate OAEP and PSS into existing interface

dignifiedquire:
This integrates #18 and #26 by expanding the use of `PaddingScheme` to capture all optional values for the different schemes.

It is not ideal, but it works well and keeps in line with the existing API for now. I think we should still continue to improve on the api and think through #34  more, but in the meantime this a version that folks can start to use, and isn't too different from the existing api

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -24,6 +24,7 @@ subtle = "2.0.0"
 simple_asn1 = "0.4"
 pem = { version = "0.7", optional = true }
 digest = { version = "0.8.0", features = ["std"] }
+sha2 = "0.8.0"
 
 [dependencies.zeroize]
 version = "1.1.0"
@@ -43,7 +44,6 @@ serde_test = "1.0.89"
 rand_xorshift = "0.2.0"
 pem = "0.7"
 sha-1 = "0.8.1"
-sha2 = "0.8.0"
 sha3 = "0.8.1"
 
 [[bench]]
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -21,6 +21,8 @@ rand = "0.7.0"
 byteorder = "1.3.1"
 failure = "0.1.5"
 subtle = "2.0.0"
+simple_asn1 = "0.4"
+pem = { version = "0.7", optional = true }
 
 [dependencies.zeroize]
 version = "1.1.0"
@@ -39,7 +41,7 @@ sha2 = "0.8.0"
 hex = "0.4.0"
 serde_test = "1.0.89"
 rand_xorshift = "0.2.0"
-
+pem = "0.7"
 
 [[bench]]
 name = "key"
@@ -51,7 +53,7 @@ name = "key"
 # debug = true
 
 [features]
-default = []
+default = ["pem"]
 nightly = ["subtle/nightly", "num-bigint/nightly"]
 serde1 = ["num-bigint/serde", "serde"]
 expose-internals = []
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -19,7 +19,7 @@ num-iter = "0.1.37"
 lazy_static = "1.3.0"
 rand = "0.7.0"
 byteorder = "1.3.1"
-failure = "0.1.5"
+thiserror = "1.0.11"
 subtle = "2.0.0"
 simple_asn1 = "0.4"
 pem = { version = "0.7", optional = true }
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -22,9 +22,9 @@ byteorder = "1.3.1"
 thiserror = "1.0.11"
 subtle = "2.0.0"
 simple_asn1 = "0.4"
-pem = { version = "0.7", optional = true }
-digest = { version = "0.8.0", features = ["std"] }
-sha2 = "0.8.0"
+pem = { version = "0.8", optional = true }
+digest = { version = "0.9.0", features = ["std"] }
+sha2 = "0.9.0"
 
 [dependencies.zeroize]
 version = "1.1.0"
@@ -38,13 +38,13 @@ default-features = false
 features = ["std", "derive"]
 
 [dev-dependencies]
-base64 = "0.11.0"
+base64 = "0.12.0"
 hex = "0.4.0"
 serde_test = "1.0.89"
 rand_xorshift = "0.2.0"
-pem = "0.7"
-sha-1 = "0.8.1"
-sha3 = "0.8.1"
+pem = "0.8"
+sha-1 = "0.9.0"
+sha3 = "0.9.0"
 
 [[bench]]
 name = "key"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -28,7 +28,8 @@ pem = { version = "0.7", optional = true }
 version = "1.1.0"
 features = ["alloc", "zeroize_derive"]
 
-[dependencies.serde]
+[dependencies.serde_crate]
+package = "serde"
 optional = true
 version = "1.0.89"
 default-features = false
@@ -55,5 +56,6 @@ name = "key"
 [features]
 default = ["pem"]
 nightly = ["subtle/nightly", "num-bigint/nightly"]
-serde1 = ["num-bigint/serde", "serde"]
+serde = ["num-bigint/serde", "serde_crate"]
+serde1 = ["serde"] # deprecated
 expose-internals = []
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "rsa"
-version = "0.1.4"
+version = "0.2.0"
 authors = ["RustCrypto Developers", "dignifiedquire <dignifiedquire@gmail.com>"]
 edition = "2018"
 description = "Pure Rust RSA implementation"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -23,6 +23,7 @@ thiserror = "1.0.11"
 subtle = "2.0.0"
 simple_asn1 = "0.4"
 pem = { version = "0.7", optional = true }
+digest = { version = "0.8.0", features = ["std"] }
 
 [dependencies.zeroize]
 version = "1.1.0"
@@ -37,12 +38,13 @@ features = ["std", "derive"]
 
 [dev-dependencies]
 base64 = "0.11.0"
-sha-1 = "0.8.1"
-sha2 = "0.8.0"
 hex = "0.4.0"
 serde_test = "1.0.89"
 rand_xorshift = "0.2.0"
 pem = "0.7"
+sha-1 = "0.8.1"
+sha2 = "0.8.0"
+sha3 = "0.8.1"
 
 [[bench]]
 name = "key"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -12,18 +12,18 @@ categories = ["cryptography"]
 readme = "README.md"
 
 [dependencies]
-num-bigint = { version = "0.5", features = ["rand", "i128", "u64_digit", "prime", "zeroize"], package = "num-bigint-dig" }
+num-bigint = { version = "0.6", features = ["rand", "i128", "u64_digit", "prime", "zeroize"], package = "num-bigint-dig" }
 num-traits = "0.2.6"
 num-integer = "0.1.39"
 num-iter = "0.1.37"
 lazy_static = "1.3.0"
-rand = "0.6.5"
+rand = "0.7.0"
 byteorder = "1.3.1"
 failure = "0.1.5"
 subtle = "2.0.0"
 
 [dependencies.zeroize]
-version = "0.10.1"
+version = "1.1.0"
 features = ["alloc", "zeroize_derive"]
 
 [dependencies.serde]
@@ -33,11 +33,12 @@ default-features = false
 features = ["std", "derive"]
 
 [dev-dependencies]
-base64 = "0.10.1"
+base64 = "0.11.0"
 sha-1 = "0.8.1"
 sha2 = "0.8.0"
 hex = "0.4.0"
 serde_test = "1.0.89"
+rand_xorshift = "0.2.0"
 
 
 [[bench]]
--- a/benches/key.rs
+++ b/benches/key.rs
@@ -6,9 +6,7 @@ use base64;
 use num_bigint::BigUint;
 use num_traits::{FromPrimitive, Num};
 use rand::{rngs::StdRng, SeedableRng};
-use rsa::hash::Hashes;
-use rsa::padding::PaddingScheme;
-use rsa::RSAPrivateKey;
+use rsa::{Hash, PaddingScheme, RSAPrivateKey};
 use sha2::{Digest, Sha256};
 use test::Bencher;
 
@@ -33,7 +31,7 @@ fn bench_rsa_2048_pkcsv1_decrypt(b: &mut Bencher) {
     let x = base64::decode(DECRYPT_VAL).unwrap();
 
     b.iter(|| {
-        let res = priv_key.decrypt(PaddingScheme::PKCS1v15, &x).unwrap();
+        let res = priv_key.decrypt(PaddingScheme::new_pkcs1v15(), &x).unwrap();
         test::black_box(res);
     });
 }
@@ -48,8 +46,7 @@ fn bench_rsa_2048_pkcsv1_sign_blinded(b: &mut Bencher) {
         let res = priv_key
             .sign_blinded(
                 &mut rng,
-                PaddingScheme::PKCS1v15,
-                Some(&Hashes::SHA2_256),
+                PaddingScheme::new_pkcs1v15_with_hash(Hash::SHA2_256),
                 &digest,
             )
             .unwrap();
--- a/benches/key.rs
+++ b/benches/key.rs
@@ -31,7 +31,9 @@ fn bench_rsa_2048_pkcsv1_decrypt(b: &mut Bencher) {
     let x = base64::decode(DECRYPT_VAL).unwrap();
 
     b.iter(|| {
-        let res = priv_key.decrypt(PaddingScheme::new_pkcs1v15(), &x).unwrap();
+        let res = priv_key
+            .decrypt(PaddingScheme::new_pkcs1v15_encrypt(), &x)
+            .unwrap();
         test::black_box(res);
     });
 }
@@ -46,7 +48,7 @@ fn bench_rsa_2048_pkcsv1_sign_blinded(b: &mut Bencher) {
         let res = priv_key
             .sign_blinded(
                 &mut rng,
-                PaddingScheme::new_pkcs1v15_with_hash(Hash::SHA2_256),
+                PaddingScheme::new_pkcs1v15_sign(Some(Hash::SHA2_256)),
                 &digest,
             )
             .unwrap();
--- a/src/algorithms.rs
+++ b/src/algorithms.rs
@@ -110,3 +110,13 @@ pub fn generate_multi_prime_key<R: Rng>(
         primes,
     ))
 }
+
+#[inline]
+pub fn copy_with_left_pad(dest: &mut [u8], src: &[u8]) {
+    // left pad with zeros
+    let padding_bytes = dest.len() - src.len();
+    for el in dest.iter_mut().take(padding_bytes) {
+        *el = 0;
+    }
+    dest[padding_bytes..].copy_from_slice(src);
+}
--- a/src/algorithms.rs
+++ b/src/algorithms.rs
@@ -1,3 +1,4 @@
+use digest::DynDigest;
 use num_bigint::traits::ModInverse;
 use num_bigint::{BigUint, RandPrime};
 use num_traits::{FromPrimitive, One, Zero};
@@ -109,4 +110,45 @@ pub fn generate_multi_prime_key<R: Rng>(
         d_final,
         primes,
     ))
+}
+
+/// Mask generation function.
+///
+/// Panics if out is larger than 2**32. This is in accordance with RFC 8017 - PKCS #1 B.2.1
+pub fn mgf1_xor(out: &mut [u8], digest: &mut dyn DynDigest, seed: &[u8]) {
+    let mut counter = [0u8; 4];
+    let mut i = 0;
+
+    const MAX_LEN: u64 = u32::MAX as u64 + 1;
+    assert!(out.len() as u64 <= MAX_LEN);
+
+    while i < out.len() {
+        let mut digest_input = vec![0u8; seed.len() + 4];
+        digest_input[0..seed.len()].copy_from_slice(seed);
+        digest_input[seed.len()..].copy_from_slice(&counter);
+
+        digest.input(digest_input.as_slice());
+        let digest_output = &*digest.result_reset();
+        let mut j = 0;
+        loop {
+            if j >= digest_output.len() || i >= out.len() {
+                break;
+            }
+
+            out[i] ^= digest_output[j];
+            j += 1;
+            i += 1;
+        }
+        inc_counter(&mut counter);
+    }
+}
+
+fn inc_counter(counter: &mut [u8; 4]) {
+    for i in (0..4).rev() {
+        counter[i] = counter[i].wrapping_add(1);
+        if counter[i] != 0 {
+            // No overflow
+            return;
+        }
+    }
 }
--- a/src/algorithms.rs
+++ b/src/algorithms.rs
@@ -119,7 +119,7 @@ pub fn mgf1_xor(out: &mut [u8], digest: &mut dyn DynDigest, seed: &[u8]) {
     let mut counter = [0u8; 4];
     let mut i = 0;
 
-    const MAX_LEN: u64 = u32::MAX as u64 + 1;
+    const MAX_LEN: u64 = std::u32::MAX as u64 + 1;
     assert!(out.len() as u64 <= MAX_LEN);
 
     while i < out.len() {
--- a/src/algorithms.rs
+++ b/src/algorithms.rs
@@ -110,13 +110,3 @@ pub fn generate_multi_prime_key<R: Rng>(
         primes,
     ))
 }
-
-#[inline]
-pub fn copy_with_left_pad(dest: &mut [u8], src: &[u8]) {
-    // left pad with zeros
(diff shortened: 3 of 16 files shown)


str4d:
Hmm, pulling in 2462f1dd56f79deb211efe7e298a9d8c2b2b8d61 from #18 goes in the opposite direction to #42. It looks like the main thing it does (remove `PublicKey` trait) is not present in the overall PR, and I can't find where that is undone, but I think it's in the refactor commit? GitHub isn't helping much here. There are other effects of that commit that are not undone (in particular, it makes public keys non-generic in `pkcs1v15`, the opposite of what I've done in #44).

How painful would it be to just drop that commit entirely? It would likely make the history clearer.

str4d:
Hmm, the change in that commit to store an `RSAPublicKey` inside `RSAPrivateKey` does make sense to me. I'd personally alter that commit to only contain the pieces we want, but that's probably more hassle than it's worth, so just reverting changes we don't want is probably fine.

dignifiedquire:
Yeah, I changed a lot of things during the merge, which is..unfortunate. So at this point I am thinking of squashing this into two commits at the end likely (1) for oaep and (2) for pss, after doing/undoing all things we want to do.

dignifiedquire:
@str4d I made the OAEP and PSS impls generic over the Key traits now

Cargo.toml
```diff
@@ -37,12 +39,12 @@ features = ["std", "derive"]
...
 [dev-dependencies]
 base64 = "0.11.0"
-sha-1 = "0.8.1"
-sha2 = "0.8.0"
 hex = "0.4.0"
 serde_test = "1.0.89"
 rand_xorshift = "0.2.0"
 pem = "0.7"
+sha-1 = "0.8.1"
+sha3 = "0.8.1"
```
tarcieri: Perhaps `sha3` (and arguably `sha-1`) should be optional? I'd say it's unusual to use any SHA-3 family function in conjunction with RSA...

dignifiedquire: these are dev dependencies

src/padding.rs
```diff
@@ -1,7 +1,83 @@
...
+}
+
+impl fmt::Debug for PaddingScheme {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        match self {
+            PaddingScheme::PKCS1v15 { ref hash } => {
+                write!(f, "PaddingScheme::PKCS1v15({:?})", hash)
+            }
+            PaddingScheme::OAEP { ref label, .. } => {
+                // TODO: How to print the digest name?
```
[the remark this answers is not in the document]
dignifiedquire: that doesn't work, the compiler is upset about more than one non auto trait :(

Cargo.toml
tarcieri: Whoops, my bad 😅

tarcieri: Whoops, my bad 😅

tarcieri: Whoops, my bad 😅

tarcieri:
Relevant to the PSS use case: final two week comment period for shipping `signature` 1.0: https://github.com/RustCrypto/traits/issues/78

src/padding.rs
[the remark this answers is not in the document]
roblabla: Ah right, it's possible to make your own trait (e.g. `trait MyDynDigest: DynDigest + Debug {}`) and then use that as a `Box<dyn MyDynDigest>`, but at this point it's too much boilerplate just for the sake of getting the digest name to show IMO.

src/pss.rs
```diff
@@ -0,0 +1,384 @@
...
+    hashed: &[u8],
+    sig: &[u8],
+    digest: &mut dyn DynDigest,
+) -> Result<()> {
+    let n_bits = pub_key.n().bits();
+    if sig.len() != (n_bits + 7) / 8 {
+        return Err(Error::Verification);
+    }
+
+    let em_bits = n_bits - 1;
```
str4d:
```suggestion
    if sig.len() != pub_key.size() {
        return Err(Error::Verification);
    }

    let em_bits = pub_key.n().bits() - 1;
```

src/pss.rs
```diff
@@ -0,0 +1,384 @@
...
+pub fn sign<T: RngCore + ?Sized, S: Rng, SK: PrivateKey>(
+    rng: &mut T,
+    blind_rng: Option<&mut S>,
+    priv_key: &SK,
+    hashed: &[u8],
+    salt_len: Option<usize>,
+    digest: &mut dyn DynDigest,
+) -> Result<Vec<u8>> {
+    let salt_len =
+        salt_len.unwrap_or_else(|| (priv_key.n().bits() + 7) / 8 - 2 - digest.output_size());
```
str4d:
```suggestion
        salt_len.unwrap_or_else(|| priv_key.size() - 2 - digest.output_size());
```

src/pss.rs
```diff
@@ -0,0 +1,384 @@
...
+    priv_key: &SK,
+    hashed: &[u8],
+    salt: &[u8],
+    digest: &mut dyn DynDigest,
+) -> Result<Vec<u8>> {
+    let n_bits = priv_key.n().bits();
+    let mut em = vec![0; ((n_bits - 1) + 7) / 8];
+    emsa_pss_encode(&mut em, hashed, n_bits - 1, salt, digest)?;
+
+    priv_key.raw_decryption_primitive(blind_rng, &em, (n_bits + 7) / 8)
```
str4d:
```suggestion
    priv_key.raw_decryption_primitive(blind_rng, &em, priv_key.size())
```

src/pss.rs
```diff
@@ -0,0 +1,384 @@
...
+fn sign_pss_with_salt<T: Rng, SK: PrivateKey>(
+    blind_rng: Option<&mut T>,
+    priv_key: &SK,
+    hashed: &[u8],
+    salt: &[u8],
+    digest: &mut dyn DynDigest,
+) -> Result<Vec<u8>> {
+    let n_bits = priv_key.n().bits();
+    let mut em = vec![0; ((n_bits - 1) + 7) / 8];
+    emsa_pss_encode(&mut em, hashed, n_bits - 1, salt, digest)?;
```
str4d:
```suggestion
    let em_bits = priv_key.n().bits() - 1;
    let mut em = vec![0; (em_bits + 7) / 8];
    emsa_pss_encode(&mut em, hashed, em_bits, salt, digest)?;
```

src/pss.rs
```diff
@@ -0,0 +1,384 @@
...
+
+    // 14. If H = H', output "consistent." Otherwise, output "inconsistent."
+    if Into::<bool>::into(h0.ct_eq(h)) {
+        Ok(())
+    } else {
+        Err(Error::Verification)
+    }
+}
+
+fn inc_counter(counter: &mut [u8]) {
```
str4d: Remove this.

src/pss.rs
```diff
@@ -0,0 +1,384 @@
...
+        counter[3] = 0u8;
+    } else {
+        counter[0] += 1;
+    }
+}
+
+/// Mask generation function
+///
+/// Will reset the Digest before returning.
+fn mgf1_xor(out: &mut [u8], digest: &mut dyn DynDigest, seed: &[u8]) {
```
str4d: Replace this with the utility function (per earlier comment).

src/oaep.rs
```diff
@@ -0,0 +1,210 @@
...
+    }
+
+    if counter[0] == u8::max_value() {
+        counter[0] = 0u8;
+        counter[1] = 0u8;
+        counter[2] = 0u8;
+        counter[3] = 0u8;
+    } else {
+        counter[0] += 1;
+    }
```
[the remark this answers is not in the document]
dignifiedquire: This function is only allowed to be called up to 2**32 times. See https://tools.ietf.org/html/rfc8017#appendix-B.2.1 for details

src/key.rs
```diff
@@ -94,6 +97,13 @@ impl Drop for RSAPrivateKey {
     }
 }
 
+impl Deref for RSAPrivateKey {
```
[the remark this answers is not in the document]
dignifiedquire: Not quite, removing this, removes the ability to call `.verify` on a private key

src/key.rs
```diff
@@ -330,6 +328,10 @@ impl<'a> PublicKeyParts for &'a RSAPrivateKey {
     fn e(&self) -> &BigUint {
         &self.e
     }
+
+    fn size(&self) -> usize {
```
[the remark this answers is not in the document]
dignifiedquire: fixed

src/key.rs
```diff
@@ -559,38 +565,61 @@ impl RSAPrivateKey {
...
     }
 
     /// Sign the given digest.
-    pub fn sign<H: Hash>(
-        &self,
-        padding: PaddingScheme,
-        hash: Option<&H>,
-        digest: &[u8],
-    ) -> Result<Vec<u8>> {
+    pub fn sign(&self, padding: PaddingScheme, input: &[u8]) -> Result<Vec<u8>> {
```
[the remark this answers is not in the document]
dignifiedquire: fixed

src/key.rs
```diff
@@ -559,38 +565,61 @@ impl RSAPrivateKey {
...
+    ///
     /// Use `rng` for blinding.
-    pub fn sign_blinded<R: Rng, H: Hash>(
+    pub fn sign_blinded<R: Rng>(
         &self,
         rng: &mut R,
         padding: PaddingScheme,
-        hash: Option<&H>,
-        digest: &[u8],
+        input: &[u8],
```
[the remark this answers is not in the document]
dignifiedquire: fixed

src/oaep.rs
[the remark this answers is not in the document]
dignifiedquire: fixed

src/oaep.rs
```diff
@@ -0,0 +1,210 @@
...
+}
+
+/// Mask generation function.
+fn mgf1_xor(out: &mut [u8], digest: &mut dyn DynDigest, seed: &[u8]) {
```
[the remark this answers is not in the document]
dignifiedquire: fixed

src/oaep.rs
```diff
@@ -0,0 +1,210 @@
...
+    let mut counter = vec![0u8; 4];
```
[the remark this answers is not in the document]
dignifiedquire: fixed

src/oaep.rs
```diff
@@ -0,0 +1,210 @@
+use rand::Rng;
+
+use digest::DynDigest;
+use subtle::{Choice, ConditionallySelectable, ConstantTimeEq};
+
+use crate::errors::{Error, Result};
+use crate::key::{self, PrivateKey, PublicKey};
+
+fn inc_counter(counter: &mut [u8]) {
```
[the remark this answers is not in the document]
dignifiedquire: fixed

src/pss.rs
dignifiedquire: fixed

src/pss.rs
dignifiedquire: fixed

src/pss.rs
dignifiedquire: fixed

src/pss.rs
dignifiedquire: fixed

src/pss.rs
dignifiedquire: fixed

src/pss.rs
dignifiedquire: fixed

src/padding.rs
```diff
@@ -1,7 +1,83 @@
...
+            PaddingScheme::PSS { ref salt_len, .. } => {
+                // TODO: How to print the digest name?
+                write!(f, "PaddingScheme::PSS(salt_len: {:?})", salt_len)
+            }
+        }
+    }
+}
+
+impl PaddingScheme {
+    pub fn new_pkcs1v15() -> Self {
```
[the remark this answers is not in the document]
dignifiedquire: added some comments on the enum variants

src/key.rs
```diff
@@ -544,13 +545,18 @@ impl RSAPrivateKey {
     pub fn decrypt(&self, padding: PaddingScheme, ciphertext: &[u8]) -> Result<Vec<u8>> {
         match padding {
             // need to pass any Rng as the type arg, so the type checker is happy, it is not actually used for anything
-            PaddingScheme::PKCS1v15 => pkcs1v15::decrypt::<ThreadRng, _>(None, self, ciphertext),
-            PaddingScheme::OAEP => unimplemented!("not yet implemented"),
+            PaddingScheme::PKCS1v15 { .. } => {
```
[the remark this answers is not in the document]
dignifiedquire: fixed

dignifiedquire:
@str4d thanks for the review, I think I addressed all important things

src/padding.rs
```diff
@@ -1,7 +1,87 @@
...
 /// Available padding schemes.
-#[derive(Debug, Clone, Copy)]
 pub enum PaddingScheme {
-    PKCS1v15,
-    OAEP,
-    PSS,
+    /// Encryption and Decryption using PKCS1v15 padding.
+    PKCS1v15Encrypt,
+    /// Sign and Verify using PKCS1v15 padding.
+    PKCS1v15Sign { hash: Option<Hash> },
```
str4d, line 13: The `hash: None` case appears to still be necessary for unpadded signatures (at least, that's what the test case in `src/pkcs1v15.rs` leads me to believe).

src/algorithms.rs
```diff
@@ -110,3 +111,44 @@ pub fn generate_multi_prime_key<R: Rng>(
...
 }
+
+/// Mask generation function.
+///
+/// Panics if out is larger than 2**32. This is in accordance with RFC 8017 - PKCS #1 B.2.1
+pub fn mgf1_xor(out: &mut [u8], digest: &mut dyn DynDigest, seed: &[u8]) {
+    let mut counter = [0u8; 4];
+    let mut i = 0;
+
+    const MAX_LEN: u64 = u32::MAX as u64 + 1;
```
str4d, line 122: This is causing the CI failure:
```suggestion
    const MAX_LEN: u64 = std::u32::MAX as u64 + 1;
```

src/oaep.rs
```diff
@@ -0,0 +1,153 @@
...
+    let hash_are_equal = db[0..h_size].ct_eq(expected_p_hash);
+
+    // The remainder of the plaintext must be zero or more 0x00, followed
+    // by 0x01, followed by the message.
+    //   looking_for_index: 1 if we are still looking for the 0x01
+    //   index: the offset of the first 0x01 byte
+    //   zero_before_one: 1 if we saw a non-zero byte before the 1
+    let mut looking_for_index = 1u8;
+    let mut index = 0u32;
+    let mut zero_before_one = 0u8;
```
str4d, line 136:
```suggestion
    let mut looking_for_index = Choice::from(1u8);
    let mut index = 0u32;
    let mut nonzero_before_one = Choice::from(0u8);
```

src/oaep.rs
```diff
@@ -0,0 +1,153 @@
...
+
+    for (i, el) in db.iter().skip(h_size).enumerate() {
+        let equals0 = el.ct_eq(&0u8);
+        let equals1 = el.ct_eq(&1u8);
+        index.conditional_assign(&(i as u32), Choice::from(looking_for_index) & equals1);
+        looking_for_index.conditional_assign(&0u8, equals1);
+        zero_before_one.conditional_assign(&1u8, Choice::from(looking_for_index) & !equals0);
```
str4d, line 143:
```suggestion
        index.conditional_assign(&(i as u32), looking_for_index & equals1);
        looking_for_index &= !equals1;
        nonzero_before_one |= looking_for_index & !equals0;
```

src/oaep.rs
```diff
@@ -0,0 +1,153 @@
...
+    }
+
+    let valid = first_byte_is_zero
+        & hash_are_equal
+        & !Choice::from(zero_before_one)
+        & !Choice::from(looking_for_index);
```
str4d, line 149:
```suggestion
        & !nonzero_before_one
        & !looking_for_index;
```

src/oaep.rs
```diff
@@ -0,0 +1,153 @@
...
+
+    Ok(out[index as usize..].to_vec())
+}
+
+/// Decrypts ciphertext using `priv_key` and blinds the operation if
+/// `rng` is given. It returns one or zero in valid that indicates whether the
+/// plaintext was correctly structured. In either case, the plaintext is
+/// returned in em so that it may be read independently of whether it was valid
+/// in order to maintain constant memory access patterns. If the plaintext was
+/// valid then index contains the index of the original message in em.
```
str4d, line 88: As currently implemented, we _don't_ maintain constant memory access patterns, because we immediately switch to a plain `Result` in `decrypt`, and I don't see any other use of `decrypt_inner`. So I don't see a reason for making `em` accessible independently of the validity.

src/oaep.rs
```diff
@@ -0,0 +1,153 @@
...
+#[inline]
+fn decrypt_inner<R: Rng, SK: PrivateKey>(
+    rng: Option<&mut R>,
+    priv_key: &SK,
+    ciphertext: &[u8],
+    digest: &mut dyn DynDigest,
+    label: Option<String>,
+) -> Result<(u8, Vec<u8>, u32)> {
```
str4d, line 96:
```suggestion
) -> Result<CtOption<(Vec<u8>, u32)>> {
```

src/oaep.rs
```diff
@@ -0,0 +1,153 @@
...
+    ciphertext: &[u8],
+    digest: &mut dyn DynDigest,
+    label: Option<String>,
+) -> Result<Vec<u8>> {
+    key::check_public(priv_key)?;
+
+    let (valid, out, index) = decrypt_inner(rng, priv_key, ciphertext, digest, label)?;
+    if valid == 0 {
+        return Err(Error::Decryption);
+    }
```
str4d, line 78:
```suggestion
    let res = decrypt_inner(rng, priv_key, ciphertext, digest, label)?;
    if res.is_none() {
        return Err(Error::Decryption);
    }
    let (out, index) = res.unwrap();
```

src/oaep.rs
```diff
@@ -0,0 +1,153 @@
...
+        zero_before_one.conditional_assign(&1u8, Choice::from(looking_for_index) & !equals0);
+    }
+
+    let valid = first_byte_is_zero
+        & hash_are_equal
+        & !Choice::from(zero_before_one)
+        & !Choice::from(looking_for_index);
+    index = u32::conditional_select(&0, &(index + 2 + (h_size * 2) as u32), valid);
+
+    Ok((valid.unwrap_u8(), em, index))
```
str4d, line 152:
```suggestion
    Ok(CtOption::new((em, index + 2 + (h_size * 2) as u32), valid))
```

src/padding.rs
dignifiedquire: Yes, which is why it is still an `Option`, not sure what you mean.

dignifiedquire:
@str4d applied the new round of CR

src/padding.rs
str4d: I was just noting this because it wasn't immediately obvious during review that the `Option<Hash>` was necessary besides enabling the prior encryption case to specify `None`.

benches/key.rs
```diff
@@ -33,7 +31,7 @@ fn bench_rsa_2048_pkcsv1_decrypt(b: &mut Bencher) {
     let x = base64::decode(DECRYPT_VAL).unwrap();
 
     b.iter(|| {
-        let res = priv_key.decrypt(PaddingScheme::PKCS1v15, &x).unwrap();
+        let res = priv_key.decrypt(PaddingScheme::new_pkcs1v15(), &x).unwrap();
```
str4d, line 34:
```suggestion
        let res = priv_key.decrypt(PaddingScheme::new_pkcs1v15_encrypt(), &x).unwrap();
```

benches/key.rs
```diff
@@ -48,8 +46,7 @@ fn bench_rsa_2048_pkcsv1_sign_blinded(b: &mut Bencher) {
         let res = priv_key
             .sign_blinded(
                 &mut rng,
-                PaddingScheme::PKCS1v15,
-                Some(&Hashes::SHA2_256),
+                PaddingScheme::new_pkcs1v15_with_hash(Hash::SHA2_256),
```
str4d, line 49:
```suggestion
                PaddingScheme::new_pkcs1v15_sign(Some(Hash::SHA2_256)),
```

src/pss.rs
```diff
@@ -0,0 +1,323 @@
...
+            //     position is "position 1") does not have hexadecimal value 0x01,
+            //     output "inconsistent" and stop.
+            for e in &db[..em_len - h_len - s_len - 2] {
+                if *e != 0x00 {
+                    return Err(Error::Verification);
+                }
+            }
+            if db[em_len - h_len - s_len - 2] != 0x01 {
+                return Err(Error::Verification);
+            }
```
str4d, line 214:
```suggestion
            let (zeroes, rest) = db.split_at(em_len - h_len - s_len - 2);
            if zeroes.iter().any(|e| *e != 0x00) || rest[0] != 0x01 {
                return Err(Error::Verification);
            }
```

src/pss.rs
```diff
@@ -0,0 +1,323 @@
...
+    // 13. Let H' = Hash(M'), an octet string of length hLen.
+    let prefix = [0u8; 8];
+
+    hash.input(&prefix[..]);
+    hash.input(m_hash);
+    hash.input(salt);
+    let h0 = hash.result_reset();
+
+    // 14. If H = H', output "consistent." Otherwise, output "inconsistent."
+    if Into::<bool>::into(h0.ct_eq(h)) {
```
str4d, line 236:
```suggestion
    if h0.ct_eq(h).into() {
```

src/pss.rs
```diff
@@ -0,0 +1,323 @@
...
+    if em_len < h_len + s_len + 2 {
+        // TODO: Key size too small
+        return Err(Error::Internal);
+    }
+
+    if em.len() != em_len {
+        return Err(Error::Internal);
+    }
+
+    let (db, h) = em.split_at_mut(em_len - s_len - h_len - 2 + 1 + s_len);
```
str4d, line 97:
```suggestion
    let (db, h) = em.split_at_mut(em_len - h_len - 1);
```

src/pss.rs
```diff
@@ -0,0 +1,323 @@
...
+    let hashed = hash.result_reset();
+    h.copy_from_slice(&hashed);
+
+    // 7.  Generate an octet string PS consisting of em_len - s_len - h_len - 2
+    //     zero octets. The length of PS may be 0.
+    //
+    // 8.  Let DB = PS || 0x01 || salt; DB is an octet string of length
+    //     emLen - hLen - 1.
+    db[em_len - s_len - h_len - 2] = 0x01;
+    db[em_len - s_len - h_len - 1..].copy_from_slice(salt);
```
str4d, line 125: We're assuming here that the `em` buffer passed into `emsa_pss_encode` is already zeroed. This happens to be the case, but given that `emsa_pss_encode` is only called from one place, and this PR isn't attempting to be no-std compatible, it would be clearer if `em` was constructed and returned from this function. Non-blocking.

dignifiedquire: fixed

src/pss.rs
dignifiedquire: fixed

src/pss.rs
dignifiedquire: fixed

src/pss.rs
dignifiedquire: fixed

src/pss.rs
```diff
@@ -0,0 +1,323 @@
...
+    mgf1_xor(db, hash, &*h);
+
+    // 9.  Set the leftmost 8 * emLen - emBits bits of the leftmost octet in DB
+    //     to zero.
+    db[0] &= 0xFF >> /*uint*/(8 * em_len - em_bits);
+
+    let s_len = match s_len {
+        None => (0..=em_len - (h_len + 2))
+            .rev()
+            .try_fold(None, |state, i| match (state, db[em_len - h_len - i - 2]) {
```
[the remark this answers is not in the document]
dignifiedquire: this change results in failures, haven't debugged why, gonna leave the logic as is for now

src/pss.rs
```diff
@@ -0,0 +1,323 @@
...
+    //
+    // 2. Let mHash = Hash(M), an octet string of length hLen
+    let h_len = hash.output_size();
+    if m_hash.len() != h_len {
+        return Err(Error::Verification);
+    }
+
+    // 3. If emLen < hLen + sLen + 2, output "inconsistent" and stop.
+    let em_len = em.len(); //(em_bits + 7) / 8;
+    if em_len < h_len + 2 {
```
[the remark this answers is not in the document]
dignifiedquire: fixed

src/pss.rs
```diff
@@ -0,0 +1,323 @@
...
+    // 4. If the rightmost octet of EM does not have hexadecimal value
+    //    0xbc, output "inconsistent" and stop.
+    if em[em.len() - 1] != 0xBC {
+        return Err(Error::Verification);
+    }
+
+    // 5. Let maskedDB be the leftmost emLen - hLen - 1 octets of EM, and
+    //    let H be the next hLen octets.
+    let (db, h) = em.split_at_mut(em_len - h_len - 1);
+    let h = &mut h[..(em_len - 1) - (em_len - h_len - 1)];
```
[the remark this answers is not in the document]
dignifiedquire: fixed

src/pss.rs
```diff
@@ -0,0 +1,323 @@
...
+    let s_len = match s_len {
+        None => (0..=em_len - (h_len + 2))
+            .rev()
+            .try_fold(None, |state, i| match (state, db[em_len - h_len - i - 2]) {
+                (Some(i), _) => Ok(Some(i)),
+                (_, 1) => Ok(Some(i)),
+                (_, 0) => Ok(None),
+                _ => Err(Error::Verification),
+            })?
+            .ok_or(Error::Verification)?,
```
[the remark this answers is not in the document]
dignifiedquire: I don't know  @roblabla I believe you wrote this code initially, do you have any answer to this?

PR #48, opened 2020-04-12
Small documentation test for key generation

Parth:
Migrating a project from openssl to this library, going to add these as I see proceed. 

Let me know if they're not needed / welcome!


dignifiedquire:
this is unfortunately very slow in debug mode to run, could you mark it as `norun`, so it doesn't slow down things too much?

dignifiedquire:
Closing for now

PR #54, opened 2020-06-27
Add RSAPrivateKey::with_e()

npmccallum:
This allows the generation of an RSA private key with a specified exponent.


aloucks:
Why not `with_exponent` ?

dignifiedquire:
fixed in #56

PR #56, opened 2020-07-29
feat: allow for custom public exponent value in keygen

kubkon:
This commit adds a function to `rsa::algorithms` called `generate_multi_prime_key_with_exp` which allows the caller to specify a custom value for the public key exponent.

This commit also adds a convenience routine to `rsa::RSAPrivateKey` called `new_with_exp` which allows the caller to specify the custom value for the public key exponent as part of `rsa::RSAPrivateKey` constructor.

Exposing the public key exponent matches an OpenSSL call `openssl::rsa::generate_with_e` which is useful in certain settings such when generating the signing keys for SGX enclaves.

Lemme know what y'all think about this change! Oh, and if you feel the PR could be improved upon or some bits are missing, lemme know and I'll be happy to iterate through with you until we get something that's mergeable!

### DIFF
--- a/src/algorithms.rs
+++ b/src/algorithms.rs
@@ -10,21 +10,46 @@ use crate::key::RSAPrivateKey;
 /// Default exponent for RSA keys.
 const EXP: u64 = 65537;
 
-// Generates a multi-prime RSA keypair of the given bit
-// size and the given random source, as suggested in [1]. Although the public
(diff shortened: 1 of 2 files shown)


kubkon:
Oh oops, this seems like a duplicate of #54. Lemme know whether we should close this one in favour of #54, or how you'd like to proceed.

kubkon:
cc @npmccallum @dignifiedquire @tarcieri

PR #59, opened 2020-08-18
chore: add Dependabot

newpavlov:
This bot helps to keep dependencies up to date.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,7 @@
+version: 2
+updates:
+- package-ecosystem: cargo
+  directory: "/"
+  schedule:
+    interval: daily
+  open-pull-requests-limit: 10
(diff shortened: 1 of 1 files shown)


dignifiedquire:
Please not, I really disklike dependabots PRs it generates a really unfortunate and unhelpful commit history in my experience.

newpavlov:
Yeah, it also can be quite annoying. :) If we had a bot which would simply notify about outdated dependencies (e.g. via issues), I would use it instead.

tarcieri:
One way to reduce dependabot churn is to turn down its alerting interval (e.g. from `daily` to `monthly`)

Also you can just use it for notifications... as soon as the relevant crates have been updated in any capacity it will auto-close obsolete PRs (I personally use it this way and find it rather helpful in that respect)

dignifiedquire:
If we can set it to monthly I would be fine. I just don't want to end up with this constant flow of notifications, for both me and others who watch the repo, whith very little signal quality.

dignifiedquire (commented):
thanks, lets try it

Issue #29, opened 2019-09-29
Slow key generation


kaimast:
Hi all,

I have the following code snippet. 

```rust
use rsa::RSAPrivateKey;
use rand::thread_rng;

let mut rng = thread_rng();
let private = RSAPrivateKey::new(&mut rng, 2048).unwrap();    
```

It sometimes takes about a second to execute, but most of the time it takes tens of seconds. Am I doing something wrong? Is there a different RNG that is faster?

Thanks!

r10s:
if you're running with the `--release` flag, things should be faster (typically few seconds).

but if you leave out this flag, things may take some tens of seconds.

at least this is what i've noticed for delta chat.

kaimast:
Thanks for the feedback!

I can confirm that it is much faster in release mode. It should not be that slow in with debug symbols enabled though.

Or is this a rust specific problem that debug bulids are much much slower?

r10s:
> Thanks for the feedback!

np

> Or is this a rust specific problem that debug bulids are much much slower?

not sure about rust, but eg, e.g https://github.com/deltachat/deltachat-core/issues/376#issuecomment-428734213 we encountered debug encryption being up to 20 times slower than release one - on c with openssl (not sure if this can be tweaked, we switched to rust then ;)

newpavlov:
@kaimast 
[`dev`](https://doc.rust-lang.org/cargo/reference/manifest.html#the-profile-sections) profile uses the lowest optimization level by default, so it's not only debug symbols, i.e. you are measuring performance for binary without any optimizations whatsoever.

olback:
I also ran into this problem. Turns out the Rust compiler is SUPER EFFICIENT when compiling a release. Generating a 2048 bit key in debug mode took arround 5 minutes, meanwhile in a release build it takes about 1 second.

dignifiedquire:
Closing, as this is expected behaviour

kaimast:
Would be great to keep this open. Key generation still takes several seconds in release mode for me. Looks like it's at least an order of magnitude slower than that of openssh. I am sure there are ways to improve upon this.

PR #61, opened 2020-09-16
docs(readme): add Note About Key Generation Speed

zicklag:
This helps address #29 a little from a documentation perspective.

### DIFF
--- a/README.md
+++ b/README.md
@@ -26,7 +26,7 @@ let dec_data = priv_key.decrypt(PaddingScheme::new_pkcs1v15(), &enc_data).expect
 assert_eq!(&data[..], &dec_data[..]);
 ```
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -26,6 +26,12 @@ let dec_data = priv_key.decrypt(PaddingScheme::new_pkcs1v15(), &enc_data).expect
 assert_eq!(&data[..], &dec_data[..]);
 ```
 
+> **Note:** If you encounter unusually slow key generation time while using `RSAPrivateKey::new` you can try to compile in release mode or add the following to your `cargo.toml`. Key generation is much faster when building with higher optimization levels, but this will increase the compile time a bit.
```
dignifiedquire, line 29: typo `Cargo.toml`

zicklag: Fixed. :+1:

PR #64, opened 2020-09-21
feat: improve OpenSSL compatibility for encoding and decoding keys

### LINKED ISSUE #63: Encoded PEM Files Not Compatible With OpenSSL or SSH
I am getting the following error when trying to decode a private key file encoded with the master version of this crate:

```bash
$ openssl rsa -in id_rsa -text
unable to load Private Key
140237402744128:error:0D078079:asn1 encoding routines:asn1_item_embed_d2i:field missing:../crypto/asn1/tasn_dec.c:406:Field=dmp1, Type=RSAPrivateKey
140237402744128:error:04096004:rsa routines:rsa_priv_decode:RSA lib:../crypto/rsa/rsa_ameth.c:183:
140237402744128:error:0606F091:digital envelope routines:EVP_PKCS82PKEY:private key decode error:../crypto/evp/evp_pkey.c:44:
140237402744128:error:0907B00D:PEM routines:PEM_read_bio_PrivateKey:ASN1 lib:../crypto/pem/pem_pkey.c:88:
```

My program looks like this:

```rust
    let mut rng = rand::thread_rng();
    let private = rsa::RSAPrivateKey::new(&mut rng, 4096).unwrap();
    let public = rsa::RSAPublicKey::from(&private);

    use rsa::PrivateKeyPemEncoding;
    std::fs::write("id_rsa", private.to_pem_pkcs8().unwrap()).unwrap();
    use rsa::PublicKeyPemEncoding;
    std::fs::write("id_rsa.pub", public.to_pem_pkcs8().unwrap()).unwrap();

    let key_data = std::fs::read_to_string("./id_rsa").unwrap();
    let key_data = rsa::pem::parse(&key_data).unwrap();
    let key = rsa::RSAPrivateKey::try_from(key_data).unwrap();
    assert_eq!(key, private);
    dbg!(key);
```

I create a key, encode it, and then read it just to make sure it survives the round trip. The round trip survives fine and all that, but for some reason OpenSSL doesn't like it. It doesn't like the public key either:

```
$ openssl rsa -in id_rsa.pub -text -pubin
unable to load Public Key
140275199583552:error:0D0BD0DC:asn1 encoding routines:c2i_ASN1_BIT_STRING:invalid bit string bits left:../crypto/asn1/a_bitstr.c:137:
140275199583552:error:0D08303A:asn1 encoding routines:asn1_template_noexp_d2i:nested asn1 error:../crypto/asn1/tasn_dec.c:627:Field=public_key, Type=X509_PUBKEY
140275199583552:error:0906700D:PEM routines:PEM_ASN1_read_bio:ASN1 lib:../crypto/pem/pem_oth.c:33:
```

If I change the bits to 512 the public key will read fine but the private key still doesn't:

```bash
$ openssl rsa -in id_rsa.pub -text -pubin
RSA Public-Key: (512 bit)
Modulus:
    00:c1:31:44:3d:2e:d6:3e:43:35:f4:ae:bc:29:24:
    c3:2b:66:d1:bc:e5:e0:9b:30:d6:fa:91:66:25:9a:
    a0:9e:0a:df:eb:f4:62:c3:f8:28:af:a0:ed:f8:48:
    e3:1b:e8:5f:24:45:b8:8c:a3:5e:d9:6e:e7:54:7d:
    92:11:12:74:07
Exponent: 65536 (0x10000)
writing RSA key
-----BEGIN PUBLIC KEY-----
MFwwDQYJKoZIhvcNAQEBBQADSwAwSAJBAMExRD0u1j5DNfSuvCkkwytm0bzl4Jsw
1vqRZiWaoJ4K3+v0YsP4KK+g7fhI4xvoXyRFuIyjXtlu51R9khESdAcCAwEAAA==
-----END PUBLIC KEY-----
```

```bash
$ unable to load Private Key
140639017108800:error:0D078079:asn1 encoding routines:asn1_item_embed_d2i:field missing:../crypto/asn1/tasn_dec.c:406:Field=dmp1, Type=RSAPrivateKey
140639017108800:error:04096004:rsa routines:rsa_priv_decode:RSA lib:../crypto/rsa/rsa_ameth.c:183:
140639017108800:error:0606F091:digital envelope routines:EVP_PKCS82PKEY:private key decode error:../crypto/evp/evp_pkey.c:44:
140639017108800:error:0907B00D:PEM routines:PEM_read_bio_PrivateKey:ASN1 lib:../crypto/pem/pem_pkey.c:88:
```

Any idea what's wrong?

smallglitch:
Reasons for the incompatibility are stated in [this comment](https://github.com/RustCrypto/RSA/issues/63#issuecomment-696336493)

Changes were made according to [RFC 3447; Appendix A.1.2](https://tools.ietf.org/html/rfc3447#appendix-A.1.2)

Closes #63

### DIFF
--- a/src/encode.rs
+++ b/src/encode.rs
@@ -17,6 +17,9 @@ const DEFAULT_ENCODING_CONFIG: EncodeConfig = EncodeConfig {
 };
 
 #[cfg(feature = "pem")]
+/// Trait for encoding the private key in the PEM format
+/// 
+/// Important: Encoding multi prime keys isn't supported. See [RustCrypto/RSA#66](https://github.com/RustCrypto/RSA/issues/66) for more info
 pub trait PrivateKeyPemEncoding: PrivateKeyEncoding {
     const PKCS1_HEADER: &'static str;
     const PKCS8_HEADER: &'static str = "PRIVATE KEY";
@@ -25,8 +28,6 @@ pub trait PrivateKeyPemEncoding: PrivateKeyEncoding {
     ///
     /// Encodes the key with the header:
     /// `-----BEGIN <name> PRIVATE KEY-----`
-    ///
-    /// Important: Encoding multi prime keys isn't supported. See [RustCrypto/RSA#66](https://github.com/RustCrypto/RSA/issues/66) for more info 
     ///
     /// # Example
     ///
@@ -45,8 +46,6 @@ pub trait PrivateKeyPemEncoding: PrivateKeyEncoding {
     }
 
     /// Converts a Private key into `PKCS1` encoded bytes in pem format with encoding config.
-    ///
-    /// Important: Encoding multi prime keys isn't supported. See [RustCrypto/RSA#66](https://github.com/RustCrypto/RSA/issues/66) for more info 
     ///
     /// # Example
     /// ```
@@ -74,8 +73,6 @@ pub trait PrivateKeyPemEncoding: PrivateKeyEncoding {
     /// Encodes the key with the header:
     /// `-----BEGIN PRIVATE KEY-----`
     ///
-    /// Important: Encoding multi prime keys isn't supported. See [RustCrypto/RSA#66](https://github.com/RustCrypto/RSA/issues/66) for more info 
-    ///
     /// # Example
     ///
     /// ```
@@ -93,8 +90,6 @@ pub trait PrivateKeyPemEncoding: PrivateKeyEncoding {
     }
 
     /// Converts a Private key into `PKCS8` encoded bytes in pem format with encoding config.
-    ///
-    /// Important: Encoding multi prime keys isn't supported. See [RustCrypto/RSA#66](https://github.com/RustCrypto/RSA/issues/66) for more info 
     ///
     /// # Example
     /// ```
@@ -231,15 +226,16 @@ fn to_bigint(value: &crate::BigUint) -> simple_asn1::BigInt {
     simple_asn1::BigInt::from_signed_bytes_le(&value.to_bigint().unwrap().to_signed_bytes_le())
 }
 
+/// Trait for encoding the private key in the PKCS#1/PKCS#8 format
+/// 
+/// Important: Encoding multi prime keys isn't supported. See [RustCrypto/RSA#66](https://github.com/RustCrypto/RSA/issues/66) for more info
 pub trait PrivateKeyEncoding {
     /// Encodes a Private key to into `PKCS1` bytes.
     ///
     /// This data will be `base64` encoded which would be used
     /// following a `-----BEGIN <name> PRIVATE KEY-----` header.
     ///
-    /// <https://tls.mbed.org/kb/cryptography/asn1-key-structures-in-der-and-pem>
-    ///
-    /// Important: Encoding multi prime keys isn't supported. See [RustCrypto/RSA#66](https://github.com/RustCrypto/RSA/issues/66) for more info 
+    /// <https://tls.mbed.org/kb/cryptography/asn1-key-structures-in-der-and-pem> 
     fn to_pkcs1(&self) -> Result<Vec<u8>>;
 
     /// Encodes a Private key to into `PKCS8` bytes.
@@ -248,8 +244,6 @@ pub trait PrivateKeyEncoding {
     /// following a `-----BEGIN PRIVATE KEY-----` header.
     ///
     /// <https://tls.mbed.org/kb/cryptography/asn1-key-structures-in-der-and-pem>
-    ///
-    /// Important: Encoding multi prime keys isn't supported. See [RustCrypto/RSA#66](https://github.com/RustCrypto/RSA/issues/66) for more info 
     fn to_pkcs8(&self) -> Result<Vec<u8>>;
 }
 
--- a/src/encode.rs
+++ b/src/encode.rs
@@ -228,6 +228,8 @@ pub trait PrivateKeyEncoding {
     /// This data will be `base64` encoded which would be used
     /// following a `-----BEGIN <name> PRIVATE KEY-----` header.
     ///
+    /// Important: Encoding multi prime keys isn't supported
+    ///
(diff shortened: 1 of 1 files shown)


tarcieri:
The failures are unrelated to this PR and appear to be due to `subtle` bumping MSRV.

I opened a separate issue for that: #65

dignifiedquire:
thanks, it definitely makes sense to stay compatible with openssl. lets get our ci game fixed and then we can land this

src/encode.rs
```diff
@@ -300,7 +316,7 @@ pub trait PublicKeyEncoding: PublicKey {
         let alg = ASN1Block::Sequence(0, vec![oid]);
 
         let bz = self.to_pkcs1()?;
-        let octet_string = ASN1Block::BitString(0, bz.len(), bz);
+        let octet_string = ASN1Block::BitString(0, bz.len() * 8, bz);
```
zicklag, line 319: Just a thought, to make the intent of this multiplication by 8 more clear we could do:

```suggestion
        let octet_string = ASN1Block::BitString(0, bz.len() * std::mem::size_of::<u8>(), bz)
```

Even though everybody knows the size of a byte at least it is lest mysterious than just `8`.

zicklag (changes requested):
Just one thought on readability.

src/encode.rs
tarcieri: `std::mem::size_of::<u8>()` is 1, not 8 (it returns bytes, not bits)

zicklag: :man_facepalming: 

Whoops, nevermind, LOL.

zicklag (approved):

smallglitch:
I've been looking at the RFC again and I think multi prime keys are still not getting correctly encoded (the additional primes are not getting encoded in a separate sequence)
I think I just haven't noticed any issues because I've been testing the code with "regular" keys

smallglitch:
I'm out of ideas. This is the code I try to use to encode the additional primes

```rust
        // Encode other primes
        if self.primes.len() > 2 {
            let mut other_prime_infos = Vec::new();

            for (index, p) in self.primes.iter().enumerate().skip(2) {
                // Initialize info with encoded prime
                let mut info = vec![ASN1Block::Integer(0, to_bigint(p))];

                // Encode exponent
                let exponent = &self.d % (p - 1u8);
                info.push(ASN1Block::Integer(0, to_bigint(&exponent)));

                // Encode coefficient
                let coefficient = self
                    .primes
                    .iter()
                    .take(index)
                    .product::<BigUint>()
                    .mod_inverse(p)
                    .ok_or(Error::InvalidModulus)?;
                info.push(ASN1Block::Integer(
                    0,
                    BigInt::from_signed_bytes_le(&coefficient.to_signed_bytes_le()),
                ));

                other_prime_infos.push(ASN1Block::Sequence(0, info));
            }

            blocks.push(ASN1Block::Sequence(0, other_prime_infos));
        }
```

When testing the resulting PEM however, OpenSSL errors out with: `RSA key error: invalid multi prime key`
Any ideas?

Relevant part of the RFC:
```
   o  otherPrimeInfos contains the information for the additional primes
      r_3, ..., r_u, in order.  It SHALL be omitted if version is 0 and
      SHALL contain at least one instance of OtherPrimeInfo if version
      is 1.

            OtherPrimeInfos ::= SEQUENCE SIZE(1..MAX) OF OtherPrimeInfo

            OtherPrimeInfo ::= SEQUENCE {
                prime             INTEGER,  -- ri
                exponent          INTEGER,  -- di
                coefficient       INTEGER   -- ti
            }

   The fields of type OtherPrimeInfo have the following meanings:

   o  prime is a prime factor r_i of n, where i >= 3.

   o  exponent is d_i = d mod (r_i - 1).

   o  coefficient is the CRT coefficient t_i = (r_1 * r_2 * ... *
      r_(i-1))^(-1) mod r_i.
```

tarcieri:
A side/meta-note, but IMO multi-prime RSA has generally gone out-of-favor given some semi-recent attacks:

https://eprint.iacr.org/2015/1137.pdf

Is it really worth supporting? Does anyone actually want to use it?

I think if people want improved performance, there's a lot of low-hanging fruit that could benefit everyone before pursuing multi-prime RSA, and without a (potentially increasingly shaky) security tradeoff.

zicklag:
> Does anyone actually want to use it?

I don't need it in my case. I've successfully used the initial patch from this PR to implement generation of SSH keys so that worked for me.

> Is it really worth supporting?

It seems like the extra threat of attack might not be worth it. If there is a problem right now were we cannot encode PEM keys with multiple primes I think we could reasonably just document it, return a result, and maybe even put a note in the docs like "If you need multi-prime key encoding open an issue to let us know that there is a need for it", or something like that.

smallglitch:
> I think we could reasonably just document it, return a result, and maybe even put a note in the docs like "If you need multi-prime key encoding open an issue to let us know that there is a need for it", or something like that

Sounds good, I'll return an `EncodingError` when the key is a multi prime and add an disclaimer to the docs. 
Something like:
```
Important: Encoding multi prime keys isn't supported
```

tarcieri:
FYI, I made a tracking ticket for multi-prime RSA discussion you can reference in an error message: RustCrypto/RSA#66

smallglitch:
Now that the CI issue, it should be possible to merge this?  
I'm still not sure what's the best solution to mention the multi-prime issue though 

Right now every function that has something to do with encoding the key has the disclaimer, which is a bit *too much* IMO.  
Adding the disclaimer to the traits itself however shows the disclaimer next to the traits directly on the main page of the documentation 

image

Both options don't feel like an ideal solution 

Opinions?

dignifiedquire:
If you could add a short description to the trait, like this

```rust
/// PrivateKeyEncoding is a great trait.
///
/// Important: Encodingi multi prime keys..
```

then it should not show up on the main page I think

smallglitch:
@dignifiedquire That worked! Thank you

dignifiedquire (approved):

dignifiedquire:
Thank you for seeing this through @smallglitch!

Issue #65, opened 2020-09-21
subtle 2.3 MSRV is 1.41+


tarcieri:
subtle 2.3's MSRV is 1.41+. This is causing the build to break, since CI's MSRV is 1.36:

https://github.com/RustCrypto/RSA/pull/64/checks?check_run_id=1146119487

@dignifiedquire any preferences on how to address this? Seems we either need to lock to earlier releases of `subtle` or bump MSRV (personally I'd prefer the latter)

dignifiedquire:
This is unfortunate. I would suggest doing a patch release pinning the version, and then a breaking release withe the updated version and new msrv.

tarcieri:
#71 pins `subtle` and gets CI passing again.

dignifiedquire:
fixed on `master`

PR #67, opened 2020-09-27
fix: pin subtle to 2.2.3 to avoid MSRV downgrades

dignifiedquire:
Ref #65


tarcieri:
Got these changes merged and CI passing again in #71

PR #70, opened 2020-11-24
Fix variable name in README

fasterthanlime:
The private key is assigned to `priv_key` but then `&private_key` is passed to `RSAPublicKey::from`.


tarcieri:
Looks like MSRV of a dependency (`subtle`?) changed

PR #71, opened 2020-11-24
fix: pin subtle to 2.1.1 to avoid MSRV downgrades

tarcieri:


tarcieri:
This at least gets CI passing again.

dignifiedquire:
thanks

PR #74, opened 2020-12-02
feat: bump dependencies and MSRV to 1.44 accordingly

dignifiedquire:


newpavlov:
What MSRV policy do you plan to follow? In other repositories we follow quite strict policy according to which MSRV bump is considered a breaking change, but it's relatively unpopular choice. Either way, I think it's worth to add a comment about MSRV policy in the readme.

dignifiedquire:
Yes, I do consider bumping the msrv a breaking change, good call on making this explicit in the readme.

dignifiedquire:
(so these changes will be released as `0.4.0`)

Issue #76, opened 2021-01-01
Suggestions on using rsa and elliptic-curve dependencies (failing on subtle)


markuskobler:
When trying to include both `rsa` and `elliptic-curve` dependencies you get a version issues with caused by `subtle` being pinned to `=2.1.1` caused by #71. What is the best path forward here to better for me to also fork and downgrade `elliptic-curve` or upgrade it for `rsa`?

```
error: failed to select a version for `subtle`.
    ... required by package `elliptic-curve v0.8.4`
    ... which is depended on by `ecdsa v0.10.2`
    ... which is depended on by `p256 v0.7.1`
...
versions that meet the requirements `=2.4.0` are: 2.4.0

all possible versions conflict with previously selected packages.

  previously selected package `subtle v2.1.1 (https://github.com/dalek-cryptography/subtle.git?tag=2.1.1#d12fc92b)`
    ... which is depended on by `rsa v0.3.0 (https://github.com/RustCrypto/RSA.git?branch=master#1d8ff05d)`
...
```

tarcieri:
Opened #77 to try to address this

tarcieri:
Looks like this was addressed in #74 and should no longer be a problem once there's another release including those changes.

markuskobler:
👍 thanks

PR #77, opened 2021-01-01
Bump `subtle` dependency to ^2.2

tarcieri:


tarcieri:
It appears `subtle` is MSRV 1.41 so the tests are failing.

I'd suggest bumping MSRV.

dignifiedquire:
I merged #74 so this should be good to go on the MSRV side now

tarcieri:
Looks like `subtle` is no-longer pinned to `=2.1.1` on HEAD, so I think #76 is already addressed (which is all I was trying to do in this PR).

PR #90, opened 2021-04-25
fix: compile when the `pem` feature is disabled

doy:


tarcieri (approved):
@doy it'd also be good to update the GitHub Actions workflow to test the feature combination that was previously failing in order to prevent regressions

doy:
yeah, that sounds like a good idea - added

dignifiedquire (approved):
thanks

PR #92, opened 2021-04-29
add NULL parameter to AlgorithmIdentifier of encoded keys

### LINKED ISSUE #91: Encoded keys missing required NULL parameter in AlgorithmIdentifier
As per the RFC 8017 [1], the NULL parameter SHALL be present if the OID in AlgorithmIdentifier is `rsaEncryption`. This crate currently omits the required NULL parameter.

From the RFC:

    The object identifier rsaEncryption identifies RSA public and private
    keys as defined in Appendices A.1.1 and A.1.2.  The parameters field
    has associated with this OID in a value of type AlgorithmIdentifier
    SHALL have a value of type NULL.

and:

```
When rsaEncryption is used in an AlgorithmIdentifier,
the parameters MUST be present and MUST be NULL.
```

[1] https://tools.ietf.org/html/rfc8017#appendix-A

clenimar:
as per the RFC 8017 [1], the NULL parameter MUST be present if the OID in
AlgorithmIdentifier is `rsaEncryption`. this commit adds an explicit
NULL to the AlgorithmIdentifier so that RSA keys encoded by this crate
are compliant to the RFC:

> The object identifier rsaEncryption identifies RSA public and private
> keys as defined in Appendices A.1.1 and A.1.2.  The parameters field
> has associated with this OID in a value of type AlgorithmIdentifier
> SHALL have a value of type NULL.

[1] https://tools.ietf.org/html/rfc8017#appendix-A

fixes #91

### DIFF
--- a/src/encode.rs
+++ b/src/encode.rs
@@ -298,7 +298,7 @@ impl PrivateKeyEncoding for RSAPrivateKey {
     fn to_pkcs8(&self) -> Result<Vec<u8>> {
         let version = ASN1Block::Integer(0, to_bigint(&BigUint::zero()));
         let oid = ASN1Block::ObjectIdentifier(0, rsa_oid());
-        let alg = ASN1Block::Sequence(0, vec![oid]);
+        let alg = ASN1Block::Sequence(0, vec![oid, ASN1Block::Null(0)]);
         let octet_string = ASN1Block::OctetString(0, self.to_pkcs1()?);
         let blocks = vec![version, alg, octet_string];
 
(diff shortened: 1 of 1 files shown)


tarcieri:
Yes, it should. [Note that this is correctly handled (or at east, easily handled correctly) by the `pkcs8` crate](https://github.com/RustCrypto/utils/blob/master/pkcs8/tests/private_key.rs#L81-L91).

I've also partially written (but have not yet published) [a `pkcs1` crate as well](https://github.com/RustCrypto/utils/blob/master/pkcs8/tests/private_key.rs#L81-L91). Once I finish that up and release it I can look at migrating the `rsa` crate over to using it.

tarcieri (approved):

clenimar:
Thanks, @randombit, I completely missed it. Now fixed.

@tarcieri: I didn't integrate to pkcs8 right away because it's currently not a dependency of this crate. If you feel strongly against using the simple fix proposed here, I could look into it. Thanks, great work!

tarcieri:
The simple fix is fine and I approved this PR. @dignifiedquire generally maintains this crate though, so I was waiting for his sign-off.

dignifiedquire:
can you rebase on master to fix ci please?

dignifiedquire:
@clenimar did you verify these with openssl?

tarcieri:
@dignifiedquire I haven't verified the encoded bytes, but this looks structurally the same as a PKCS#8-encoded RSA key generated by OpenSSL

https://github.com/RustCrypto/utils/blob/master/pkcs8/tests/private_key.rs#L81-L91

est31:
@dignifiedquire I tried out this PR locally with the example from #95 and it works. So with this PR, ring can successfully import the key.

clenimar:
> @clenimar did you verify these with openssl?

@dignifiedquire yes, I did. Also been using this PR with ring, as @est31 mentioned. Rebased now.

Thanks!

Issue #94, opened 2021-05-10
Remove std feature requirement for encoding


AndreVallestero:
I would like to use `PrivateKeyPemEncoding` in my Wasm project but RSA currently cannot target Wasm with feature `std` enabled. I am able to compile and generate keys successfully with feature `std` disabled, however I am unable to encode it to a usable format for my use case (pem, pkcs1, jwk).

tarcieri:
The `pkcs8` crate is designed to provide `no_std`-friendly support for PKCS#8 and also supports PEM:

https://github.com/RustCrypto/utils/tree/master/pkcs8

We also own the `pkcs1` crate name. It should be pretty easy to define a similar/compatible `no_std` PKCS#1 encoder (using the [`der` crate](https://github.com/RustCrypto/utils/tree/master/der)).

tarcieri:
I think it's a desired goal for the project, but possibly a dup of #75

dignifiedquire:
fixed on `master`

PR #98, opened 2021-06-15
remove std requirement

OlivierLemoine:
Hi, I remove the std requirement in order to use it in WebAssembly.
Is there a reason for this restriction ?


tarcieri:
I assume by this you mean "remove std requirement for encoding", in which case this could address #94

tarcieri:
Looks like an unrelated build failure on nightly owing to a removed feature

dignifiedquire:
thanks. can you try and rebase on master and see if you can fix ci?

dignifiedquire:
Closing in favor of #104

PR #101, opened 2021-07-19
Add possibility to allow a different hash function for MGF mask function

ubamrein:
This pull request adds a new field to the OAEP padding to allow for a different hash function for the MGF mask function. This is especially needed, when trying to be compatible with other algorithms, more specifically the `AndroidKeyStore`. The `AndroidKeyStore` only supports RSA encryption with SHA256 and SHA1 for the MGF mask function.

 For compatibility the existing functions were adjusted to replicate the previous behaviour to use the same hash function.

To the OAEP-Padding  two new functions `new_oaep_with_mgf_hash` and `new_oaep_with_mgf_hash_with_labe` were added to allow specifying the MGF hash function.

### DIFF
--- a/src/key.rs
+++ b/src/key.rs
@@ -191,8 +191,8 @@ impl PublicKey for RSAPublicKey {
     fn encrypt<R: Rng>(&self, rng: &mut R, padding: PaddingScheme, msg: &[u8]) -> Result<Vec<u8>> {
         match padding {
             PaddingScheme::PKCS1v15Encrypt => pkcs1v15::encrypt(rng, self, msg),
-            PaddingScheme::OAEP { mut digest, label } => {
-                oaep::encrypt(rng, self, msg, &mut *digest, label)
+            PaddingScheme::OAEP { mut digest, mut mgf_digest, label } => {
+                oaep::encrypt(rng, self, msg, &mut *digest, &mut *mgf_digest, label)
             }
             _ => Err(Error::InvalidPaddingScheme),
         }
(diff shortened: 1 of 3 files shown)


tarcieri (approved):
Looks good to me, although I'm not sure the function name changes are required. To me just adding an explicit argument for the MGF is sufficient.

~~Clearly a breaking change, so I'll defer to @dignifiedquire before merging.~~ Whoops seems I misread, this is purely additive.

tarcieri:
Tests are breaking due to a `zeroize` MSRV bump

ubamrein:
I just tried to avoid introducing too many breaking changes. This way, old code relaying on the `new_oaep` function does not need to be changed.

ubamrein:
I'm a bit confused, as to why the version of zeroize was bumped to 1.4.0? In `RSA`'s TOML 1.1.0 is referenced, but apparently 1.4.0 is used (`cargo tree`). Is there a way to fix that? :D

tarcieri:
This repo doesn't check in `Cargo.lock` so it always uses the latest version.

I can fix it in a separate PR.

tarcieri:
@ubamrein can you rebase?

dignifiedquire (changes requested):
overall looks good, but can you add some documentation for this please, describing how it can be used?

ubamrein:
I'm not quite sure what exactly I should add, or rather, if I also should try to add documentation for the old functions/behaviour?

dignifiedquire:
> if I also should try to add documentation for the old functions/behaviour?

that would be awesome, yes

ubamrein:
I added some documentation, and referenced the standard directly. I'm not 100% sure about the phrasing though, so I would appreciate any feedback :) 

 As this PR is related to the OAEP padding, I didn't want to change stuff not related to it.

dignifiedquire (approved):
thank you!

ubamrein:
I will push some minor improvements to the documentation, so please don't merge yet.

PR #104, opened 2021-07-24
Use `pkcs1` and `pkcs8` crates; MSRV 1.51+

### LINKED ISSUE #75: `pkcs8` crate
FYI, I've just published a `pkcs8` crate which the https://github.com/RustCrypto/elliptic-curves projects are now using:

- GitHub: https://github.com/RustCrypto/utils/tree/master/pkcs8
- Crate: https://crates.io/crates/pkcs8
- Docs: https://docs.rs/pkcs8/

It's `no_std`-friendly, mostly self-contained including its own PEM parser (which depends only on `alloc` at present), and also provides [`pkcs8::FromPrivateKey`](https://docs.rs/pkcs8/latest/pkcs8/trait.FromPrivateKey.html) and [`pkcs8::FromPublicKey`](https://docs.rs/pkcs8/latest/pkcs8/trait.FromPublicKey.html) traits which wrap up PKCS#8/SPKI (respectively) decoding from BER/DER and PEM formats, including loading keys directly from the filesystem.

I think it'd be great if the `rsa` crate adopted these and will work on a PR if I have time, unless someone else is interested.

One nice benefit would be making it easier for people to write libraries that work with e.g. either RSA(SSA) or ECDSA signatures.

tarcieri:
Closes #75

Switches the `rsa` crate to use the `pkcs1` and `pkcs8` crates from https://github.com/rustcrypto/utils

# Rationale

- **Interoperability**: PKCS#8 in particular is algorithm agnostic, and the `pkcs8` crate provides the following traits which can be used to abstract over the algorithm used by a particular key (NOTE: this PR does not yet impl these traits, but probably should)
  - [`FromPrivateKey`](https://docs.rs/pkcs8/latest/pkcs8/trait.FromPrivateKey.html)
  - [`FromPublicKey`](https://docs.rs/pkcs8/latest/pkcs8/trait.FromPublicKey.html)
  - [`ToPrivateKey`](https://docs.rs/pkcs8/latest/pkcs8/trait.ToPrivateKey.html)
  - [`ToPublicKey`](https://docs.rs/pkcs8/0latest/pkcs8/trait.ToPublicKey.html)
- **Security**: an [upcoming ACM CCS paper](https://twitter.com/JanWichelmann/status/1418532480081145857) claims to be able to extract RSA private keys from SGX using PEM parsing sidechannels alone. The new [`pem-rfc7468`](https://github.com/RustCrypto/utils/tree/master/pem-rfc7468) crate implements a constant time PEM parser which leverages the [`base64ct`](https://github.com/RustCrypto/utils/tree/master/base64ct) crate for constant-time encoding/decoding.
- **Parsimony**: by switching to crates from https://github.com/rustcrypto/utils a user of the `rsa` crate who is also leveraging any of RustCrypto's elliptic curve crates such as `p256`, `p384`, or `k256` will have fewer overall dependencies.
- **PKCS#8 Encryption**: the `pkcs8` crate has support for `ENCRYPTED PRIVATE KEY`s using the `pkcs5` crate. With this it can encrypt any PKCS#8 key under a password with scrypt-based key derivation and AES-CBC encryption (unfortunately PKCS#8 has no support for AEADs, alas)

# Notes

This is the first time I've attempted to use the `pkcs1` crate as it's brand new, so this integration is effectively taking it out for a test drive.

One of the key parts of format encoding/decoding with RSA when implementing PKCS#1 and PKCS#8 is that the latter is effectively a wrapper for the former in modern use. So therefore it seems like there should be a first-class integration between the `pkcs1` and `pkcs8` crates, namely I think the `pkcs1` crate should support an optional dependency on `pkcs8`.

While the `pkcs8` crate has some nice traits like the aforementioned `FromPrivateKey`/`ToPrivateKey`, the `pkcs1` crate does not. The `rsa` crate defines some traits for this (e.g. `PrivateKeyEncoding`, `PrivateKeyPemEncoding`) but I really feel like the traits should get hoisted up into the `pkcs1` crate.

If that were to happen, I think that the integration with the `rsa` crate could *just* be for PKCS#1 DER encoding/decoding. Concerns like PKCS#8 and PEM could be hoisted up into traits defined by the `pkcs1` crate.


tarcieri:
Sidebar:

I noticed the `rsa` crate does the following:

```rust
#![cfg_attr(not(test), no_std)]
```

An alternative to this is to have a separate integration tests directory i.e. `tests/` ala here:

https://github.com/RustCrypto/utils/tree/master/pkcs8/tests

Integration tests are linked separately from the crate you're developing and automatically get `std`, even if the crate in `src/` is `#[no_std]`, so you don't need to conditionally link `std` for testing.

There are a few other items marked `#[cfg(test)]` and I think using integration tests like that helps keep the main `src/lib.rs` clean.

Also I think it'd be nice to standardize on a set of test vectors. The ones in the `pkcs8` crate were generated by myself using recent versions of OpenSSL and I've reused them with the other crates in the utils repo as well as the https://github.com/rustcrypto/elliptic-curves repo.

newpavlov:
>An alternative to this is to have a separate integration tests directory i.e. tests/ ala here:

It's not always possible to use integration tests, e.g. if tests are for private internals of a module. But apart from that I agree that we should prefer the `tests/` folder if possible. Also a more common approach for handling optional `std` dependency is:
```rust
#![no_std]

#[cfg(features = "std")]
extern crate std;
```
We can modify it to use `#[cfg(any(features = "std", test))]`.

src/pss.rs
```diff
@@ -280,14 +280,12 @@ mod test {
             let digest = Sha1::digest(test[0].as_bytes()).to_vec();
             let sig = hex::decode(test[1]).unwrap();
 
-            let seed = SystemTime::now().duration_since(SystemTime::UNIX_EPOCH).unwrap();
+            let seed = SystemTime::now()
+                .duration_since(SystemTime::UNIX_EPOCH)
+                .unwrap();
```
newpavlov, line 285: I don't quite like non-deterministic tests... Maybe we should port them to RNGs which use a fixed seed? Also it may be worth using `rand_core` instead of `rand`.

tarcieri: I think this change was the result of my IDE applying rustfmt to the project, FWIW. That's kind of annoying, I was hoping it would work at the individual module level.

tarcieri:
> It's not always possible to use integration tests, e.g. if tests are for private internals of a module. 

Well yeah, I'd call those unit tests 😉

tarcieri:
Note: I'm going to add a set of traits to the `pkcs1` crate for encoding/decoding keys similar to the traits in the `pkcs8` crate:

https://github.com/RustCrypto/utils/pull/540

From there I'd like to add an optional `pkcs8` crate to the `pkcs1` crate, and a blanket impl of the `pkcs8` key traits for types which impl the ones from the `pkcs1` crate. That will make PKCS#8 support reusable for any types which impl the PKCS#1 traits.

tarcieri:
Note that this is going to need an MSRV bump to 1.51, so maybe it would make sense to merge #92 first, cut another release, and then merge this after cutting the next release

tarcieri:
Removing WIP/draft.

Aside from some rustdoc on how to use the traits from the `pkcs1`/`pkcs8` crates I'd say this is good to go.

It's now extensively tested with all of the RSA vectors from the `pkcs1` and `pkcs8` crates, to ensure they parse correctly and round trip as both DER and PEM.

I'd suggest merging #92 (quick fix for PKCS#8) and #105 (rustfmt) first, as re: the former it doesn't need an MSRV bump and re: the latter this PR has rustfmt applied and with it some irrelevant changes it'd be nice to get taken care of in a rustfmt-specific PR.

src/pss.rs
tarcieri: rustfmt applied in #105

tarcieri:
Calling this finished for now. I've added some preliminary documentation, although it's not quite as extensive as the documentation it's replacing. I can add more if desired.

Cargo.toml
```diff
@@ -20,9 +20,9 @@ lazy_static = { version = "1.3.0", features = ["spin_no_std"] }
...
 byteorder = { version = "1.3.1", default-features = false }
 subtle = { version = "2.1.1", default-features = false }
-simple_asn1 = { version = "0.5", optional = true }
-pem = { version = "0.8", optional = true }
 digest = { version = "0.9.0", default-features = false }
+pkcs1 = { version = "0.2.2", default-features = false }
+pkcs8 = { version = "0.7.4", default-features = false }
 
 [dependencies.zeroize]
 version = "=1.3" # zeroize 1.4 is MSRV 1.51+
```
dignifiedquire, line 28: if we bump MSRV to 1.51, lets update this as well

src/encode.rs
```diff
@@ -1,463 +0,0 @@
...
-    /// use pem::{EncodeConfig, LineEnding};
-    /// # let mut rng = OsRng;
-    /// # let bits = 2048;
-    /// # let private_key = RSAPrivateKey::new(&mut rng, bits).expect("failed to generate a key");
-    ///
-    /// let _ = private_key.to_pem_pkcs1_with_config(EncodeConfig {
-    ///     line_ending: LineEnding::CRLF,
-    /// });
-    /// ```
-    fn to_pem_pkcs1_with_config(&self, config: EncodeConfig) -> Result<String> {
```
dignifiedquire, line 65: Is there a way to get the line encoding config with the pkcs1/pkcs8 crate?

src/pss.rs
dignifiedquire: created an issue to fix the seeds https://github.com/RustCrypto/RSA/issues/108

src/encode.rs
tarcieri: Currently a TODO, although not too hard to add implementation-wise: https://github.com/RustCrypto/utils/blob/master/pem-rfc7468/src/encoder.rs#L12-L14

tarcieri: PR to add support for selecting line endings to `pem-rfc7468`:

https://github.com/RustCrypto/utils/pull/551

tarcieri: Here's a PR to add line ending customization to `pkcs1`, using the newly released `pem-rfc7468`:

https://github.com/RustCrypto/utils/pull/553/files

If that looks good I can do `pkcs8` too.

tarcieri: Went ahead and cut new releases of `pkcs1` and `pkcs8` with this feature added (0.2.3/0.7.5 respectively), and bumped them in a9de07d5313590d68353cdfb1b627f0488700377

Cargo.toml
tarcieri: Bumped MSRV in 2f2a67388137129357ab518953a8cc7fad5fc490.

I went with a less restrictive `>=1, <1.5` requirement to ensure `rsa` can be used alongside crates which are locking `=1.3`. Otherwise it wouldn't be possible to use both /cc @newpavlov

dignifiedquire (approved):
Awesome work, thank you! I love the smell of delted code and more constant time options even more.

dignifiedquire:
Released 0.4.1, so feel free to rebase & merge when you are ready.

tarcieri:
Rebased

PR #105, opened 2021-07-25
rustfmt

tarcieri:
Applies `rustfmt` with the following version:

    rustfmt 1.4.37-stable (2a3635d5 2021-05-04)

### DIFF
--- a/src/algorithms.rs
+++ b/src/algorithms.rs
@@ -1,11 +1,11 @@
+use alloc::vec;
 use digest::DynDigest;
 use num_bigint::traits::ModInverse;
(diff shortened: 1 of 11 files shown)


tarcieri:
Would also probably suggest adding a GitHub Action to check `rustfmt`, e.g.:

https://github.com/RustCrypto/elliptic-curves/blob/77e2a92/.github/workflows/workspace.yml#L29-L47

tarcieri:
Okay, this is driving me insane as rustfmt is a pretty integral part of my workflow, so I'm going to go ahead and merge this.

dignifiedquire:
thanks

PR #106, opened 2021-07-25
Remove `extern crate` directives

tarcieri:
These are no longer needed as of the 2018 edition


tarcieri:
This one seems innocuous enough

PR #109, opened 2021-07-26
docs(readme): only optimizing num-bigint-dig is enough to unlock most speedups

est31:
I saw a merge of #61 and thought that integrating the limited optimization instruction I found would be a nice idea.


dignifiedquire (commented):
good idea, thanks

PR #110, opened 2021-07-26
refactor: rename  `RSA*` to `Rsa`

### LINKED ISSUE #24: Type names
Right now this crate uses capitalized RSA in type names, e.g. `RSAPrivateKey`. But it looks like according to Rust naming conventions it should be named as `RsaPrivateKey`, see for example [`UdpSocket`](https://doc.rust-lang.org/std/net/struct.UdpSocket.html).

dignifiedquire:
This matches the rust naming convention better.

Closes #24


tarcieri (approved):
Noticed this doing the `pkcs1` integration... will be nice for this to match the `pkcs1::Rsa*` names

PR #114, opened 2021-09-30
feat: derive Hash for `RsaPublicKey`

Mallets:
It would be great for the RsaPublicKey to derive the Hash trait.
All the internal fields of RsaPublicKey already implement the Hash trait, it is enough to automatically derive.
By doing so, it becomes possible to store RsaPublicKey e.g. in an HashSet.


dignifiedquire (approved):
makes sense, thanks

PR #117, opened 2021-11-14
Add a sha1 feature to enable pkcs8/sha1

jpalathu:
Currently, the RSA crate doesn't have a feature to enable the sha1 feature in the pkcs8 crate. It would be great if we could add this feature to enable sha1 from this crate.


jpalathu:
@tarcieri any thoughts on this?

tarcieri:
@dignifiedquire generally reviews PR for and maintains the `rsa` crate

My initial thought on this is that it probably isn't worth adding a `sha1` feature for this alone. Maybe if there are strong interop reasons it might be useful e.g. as an MGF.

As an alternative you can add `pkcs8` as a direct dependency:

```toml
[dependencies]
pkcs8 = { features = ["sha1"], ... }
rsa = { ... }
```

dignifiedquire:
yeah in a similar boat, I would like to avoid adding feature flags for all features of each dependency. the snippet at the end above should do the trick for you. If not please describe the concrete use case

tarcieri:
Think we can close this then.

@jpalathu hopefully the snipped I provided works for you.

PR #124, opened 2021-12-03
refactor: remove `lazy_static` dependency

Expyron:
This also speeds up `check_public` by about 30%, as the compiler can optimize it by taking into account the constant values.

### DIFF
--- a/src/key.rs
+++ b/src/key.rs
@@ -1,10 +1,9 @@
 use alloc::vec::Vec;
 use core::ops::Deref;
-use lazy_static::lazy_static;
 use num_bigint::traits::ModInverse;
 use num_bigint::Sign::Plus;
 use num_bigint::{BigInt, BigUint};
-use num_traits::{FromPrimitive, One};
+use num_traits::{One, ToPrimitive};
 use rand::{rngs::StdRng, Rng};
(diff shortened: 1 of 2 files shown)


Expyron:
I updated this PR with a cleaner and much faster version.

src/key.rs
```diff
@@ -512,11 +509,13 @@ impl RsaPrivateKey {
 /// Check that the public key is well formed and has an exponent within acceptable bounds.
 #[inline]
 pub fn check_public(public_key: &impl PublicKeyParts) -> Result<()> {
-    if public_key.e() < &*MIN_PUB_EXPONENT {
+    let public_key: u64 = public_key.e().to_u64().ok_or(Error::PublicExponentTooLarge)?;
```
tarcieri, line 512: IMO the `to_u64` is clear enough this doesn't need a `public_key: u64` type annotation

tarcieri (approved):
One nit otherwise LGTM

src/key.rs
Expyron: Fixed.
Thank you for the review.

Expyron:
This is probably good to merge now.

dignifiedquire (approved):
great, thank you

PR #126, opened 2021-12-04
Adds example of reading/writing RSA keys to/from file

quietlychris:
Hello,

I started working with the `rsa` crate today, and it took me a while to figure out the idiomatic ways of reading/writing keys to and from .pem files. In retrospect, I can see that the re-export of the `pkcs#` crates offered those methods, but it took a while (I actually re-implemented the with `std::fs::read_to_string` and other methods before finding them). 

I figured that other people might run into a similar issue at some point, so figured I might offer an example in the repository in addition to the documentation, as suggested in #7. I've added the `cargo build --examples --all` to the  Github Actions CI, and `clippy` should be happy. Thanks for your consideration, and happy to make any changes or answer any questions!

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,4 @@
 /target
 **/*.rs.bk
 Cargo.lock
+*id_rsa*
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -61,9 +61,9 @@ jobs:
       - name: Examples
         uses: actions-rs/cargo@v1
         with:
-          use-cross: true
+          use-cross: false
           command: build 
-          args: --examples --all --target ${{ matrix.target }} --verbose ${{ matrix.args }}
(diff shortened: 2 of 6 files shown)


quietlychris:
The CI is obviously very unhappy with me at the moment--I'll spend some time soon trying to replicate the failure and hopefully fix it before pushing more commits. Sorry about this!

.gitignore
```diff
@@ -1,3 +1,4 @@
 /target
 **/*.rs.bk
 Cargo.lock
+*id_rsa*
```
aloucks, line 4: Why don't you read and write the example data inside the `target/` directory.

quietlychris: Good point, I'll change the default file location.

quietlychris:
Closing until I can make time to solve this in a separate branch

PR #135, opened 2022-01-05
feat: update to rust edition 2021

dignifiedquire:
bumps MSRV to 1.56.0

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -72,17 +72,3 @@ jobs:
           command: bench
           args: --target ${{ matrix.test-target || matrix.target }} --verbose --no-run ${{ matrix.args }}
         if: matrix.rust == 'nightly'
-
-  minimal-versions:
-    runs-on: ubuntu-latest
-    steps:
-      - uses: actions/checkout@v2
-      - uses: RustCrypto/actions/cargo-cache@master
-      - uses: actions-rs/toolchain@v1
-        with:
-          toolchain: nightly
-          override: true
-          profile: minimal
-      - uses: RustCrypto/actions/cargo-hack-install@master
-      - run: cargo update -Z minimal-versions
-      - run: cargo hack test --release --feature-powerset
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -74,6 +74,15 @@ jobs:
         if: matrix.rust == 'nightly'
 
   minimal-versions:
-    uses: RustCrypto/actions/.github/workflows/minimal-versions.yml@master
-    with:
-        working-directory: ${{ github.workflow }}
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+      - uses: RustCrypto/actions/cargo-cache@master
+      - uses: actions-rs/toolchain@v1
+        with:
+          toolchain: nightly
+          override: true
+          profile: minimal
+      - uses: RustCrypto/actions/cargo-hack-install@master
+      - run: cargo update -Z minimal-versions
+      - run: cargo hack test --release --feature-powerset
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -72,3 +72,8 @@ jobs:
           command: bench
           args: --target ${{ matrix.test-target || matrix.target }} --verbose --no-run ${{ matrix.args }}
         if: matrix.rust == 'nightly'
+
+  minimal-versions:
+    uses: RustCrypto/actions/.github/workflows/minimal-versions.yml@master
+    with:
+        working-directory: ${{ github.workflow }}
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -2,7 +2,7 @@
 name = "rsa"
 version = "0.5.0"
 authors = ["RustCrypto Developers", "dignifiedquire <dignifiedquire@gmail.com>"]
-edition = "2018"
+edition = "2021"
 description = "Pure Rust RSA implementation"
 license = "MIT OR Apache-2.0"
 documentation = "https://docs.rs/rsa"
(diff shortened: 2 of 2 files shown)


newpavlov:
It's also worth to add `rust-version = "1.56"` field as well.

dignifiedquire:
I wonder why nightly CI is unhappy now

paolobarbolini:
Wih this MSRV bump it should now be possible to relax the `zeroize` version requirement

tarcieri:
> I wonder why nightly CI is unhappy now

The `nightly` feature of the `subtle` crate is using a feature which has been stabilized in 1.56:

```
error[E0557]: feature has been removed
  --> /Users/bascule/.cargo/registry/src/github.com-1ecc6299db9ec823/subtle-2.4.0/src/lib.rs:12:42
   |
12 | #![cfg_attr(feature = "nightly", feature(external_doc))]
   |                                          ^^^^^^^^^^^^ feature has been removed
   |
   = note: use #[doc = include_str!("filename")] instead, which handles macro invocations
```

Note that the `nightly` feature of `subtle` doesn't provide much value anymore, so it might be good not to activate it, particularly since it doesn't compile on recent nightlies anymore.

tarcieri:
Wow that's really weird... almost feels like a `nightly` bug?

@dignifiedquire maybe try bumping `num-bigint-dig` to the 2021 edition first?

poliorcetics:
CI gets the following error:

```
 error[E0599]: no function or associated item named `deserialize` found for struct `Vec<_, _>` in the current scope
Error:     --> /cargo/registry/src/github.com-1ecc6299db9ec823/num-bigint-dig-0.7.0/src/biguint.rs:2582:35
     |
2582 |         let data: Vec<u32> = Vec::deserialize(deserializer)?;
     |                                   ^^^^^^^^^^^ function or associated item not found in `Vec<_, _>`
```

I'm not managing to reproduce it on `num-bigint`, tag `v0.7.0`, with all the features used by `cg +nightly build --features=i128,nightly,prime,rand,serde,u64_digit,zeroize` but compiling this PR with `cargo +nightly build --all-features` **does** have the error.

Nightly version: `cargo 1.60.0-nightly (358e79fe5 2022-01-04)`

poliorcetics:
Taking this PR and changing the edition to 2018 works, this is very probably a rustc bug due to `num-bigint(-dig)` not setting an edition in `Cargo.toml` (so defauting to 2015). This probably cannot be merged without `num-bigint` moving to 2018 or 2021 edition first.

tarcieri:
I opened a PR to bump the edition for `num-bigint-dig` to 2021: https://github.com/dignifiedquire/num-bigint/pull/34

poliorcetics:
Asked for a 0.8 update for `num-bigint-dig` so it can be updated in this PR and complete it

paolobarbolini:
num-bigint-dig 0.8 is out!
https://crates.io/crates/num-bigint-dig

dignifiedquire:
unfortunately this did not fix the rustc bug :'(

paolobarbolini:
I'm from my phone so I can't test now but could it be caused by the serde alloc feature not being enabled? 

https://github.com/dignifiedquire/num-bigint/blob/9c7fbac52c48e05cff75104f3a1278bd577b3ee3/Cargo.toml#L57

https://github.com/RustCrypto/RSA/runs/5246995762?check_suite_focus=true#step:4:270

paolobarbolini:
Confirmed by running `cargo check --no-default-features --features i128,u64_digit,prime,zeroize,serde` and  `cargo tree --edges normal --format "{p} {f}" --no-default-features --features i128,u64_digit,prime,zeroize,serde` on num-bigint-dig

newpavlov:
It could be worth to add the [minimal versions job](https://github.com/RustCrypto/hashes/blob/master/.github/workflows/sha3.yml#L48-L51) to CI.

.github/workflows/ci.yml
```diff
@@ -72,3 +72,8 @@ jobs:
           command: bench
           args: --target ${{ matrix.test-target || matrix.target }} --verbose --no-run ${{ matrix.args }}
         if: matrix.rust == 'nightly'
+
+  minimal-versions:
+    uses: RustCrypto/actions/.github/workflows/minimal-versions.yml@master
+    with:
+        working-directory: ${{ github.workflow }}
```
newpavlov, line 79: Argh, you can not use the shared config. You can use the following config instead:
```yaml
  minimal-versions:
    runs-on: ubuntu-latest
    steps:
      - uses: actions/checkout@v2
      - uses: RustCrypto/actions/cargo-cache@master
      - uses: actions-rs/toolchain@v1
        with:
          toolchain: nightly
          override: true
          profile: minimal
      - uses: RustCrypto/actions/cargo-hack-install@master
      - name: Test benches build
        run: cargo build --benches
      - run: cargo update -Z minimal-versions
      - run: cargo hack test --release --feature-powerset
```

poliorcetics:
CI improvement should be delayed to another MR, this is blocking a new release of RSA which in turn blocks the whole ecosystem that uses both RSA and another crate that has made the jump to a recent version of `hashes`.

I can take the commits except the last one and put them in another MR to move forward if needed.

tarcieri:
I tried @newpavlov's `minimal-versions` config. It looks like `num-bigint-dig` v0.8.1 is not configured with the correct versions for that to work: https://github.com/RustCrypto/RSA/runs/5497433460?check_suite_focus=true

So I agree it would make sense to get this PR merged first then circle back on making `num-bigint-dig` minimal-versions correct and then following up on `minimal-versions` CI in a separate PR.

.github/workflows/ci.yml
```diff
@@ -72,3 +72,17 @@ jobs:
...
+      - uses: actions/checkout@v2
+      - uses: RustCrypto/actions/cargo-cache@master
+      - uses: actions-rs/toolchain@v1
+        with:
+          toolchain: nightly
+          override: true
+          profile: minimal
+      - uses: RustCrypto/actions/cargo-hack-install@master
+      - run: cargo update -Z minimal-versions
+      - run: cargo hack test --release --feature-powerset
```
tarcieri, line 88:
```suggestion
```

newpavlov:
I am fine with leaving the CI changes for later, but I think the minimal versions build should be fixed before we release `rsa v0.6`.

tarcieri (approved):

tarcieri:
FYI for anyone particularly blocked here: I just cut a v0.6.0-pre release with these changes

https://crates.io/crates/rsa/0.6.0-pre

PR #139, opened 2022-01-16
Bump `pkcs1` => v0.3, `pkcs8` => v0.8; MSRV 1.56

tarcieri:
Updates the `pkcs1` and `pkcs8` crates to their latest versions.

These releases reversed the blanket impl relationship between the traits in the two respective crates.

The names of the traits have changed:
- `From*Key` => `Decode*Key` (e.g. `DecodePrivateKey`)
- `To*Key` => `Encode*Key` (e.g. `EncodePrivateKey`).

Previously there was a blanket impl of `pkcs8` traits for the `pkcs1` crates, e.g. `DecodePrivateKey` was impl'd for `DecodeRsaPrivateKey`.

This relationship has been reversed in the latest releases, and as such this commit now implements the `pkcs8` traits instead of the `pkcs1` traits (even though this internally involves decoding a PKCS#1 document as part of PKCS#8 decoding).

The reasoning is this was necessary to make it possible for traits for multiple formats to have blanket impls for the PKCS#8 traits. Concretely this was needed to allow blanket impls of the traits in the `sec1` crate for types which impl traits from the `pkcs8` crate.

Both of these crates are MSRV 1.56 and hard dependencies, so the MSRV of the `rsa` crate is accordingly bumped to 1.56.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -19,7 +19,7 @@ jobs:
     strategy:
       matrix:
         rust:
-          - 1.51.0
+          - 1.56.0
           - stable
           - nightly
         target:
@@ -36,7 +36,7 @@ jobs:
             target: x86_64-unknown-linux-gnu
             args: --no-default-features --features=std
         exclude:
-          - rust: 1.51.0
+          - rust: 1.56.0
             target: thumbv7m-none-eabi
 
     steps:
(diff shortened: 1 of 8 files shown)


tarcieri:
Note: I didn't do a 2021 edition upgrade as part of this PR since #135 is already open for this purpose

dignifiedquire (approved):
great, thank you

PR #141, opened 2022-01-25
mark key and pkcs1v15 as public to give access to verify

carl-wallace:
Enables code like the following to verify signatures using keys extracted from a certificate.

```
            let rsa = RsaPublicKey::from_public_key_der(&enc_spki);
            if let Ok(rsa) = rsa {
                let x = rsa.verify(ps, hash_to_verify, signature);

```

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -182,13 +182,13 @@ pub mod padding;
 mod encoding;
 #[cfg(feature = "alloc")]
 #[cfg_attr(docsrs, doc(cfg(feature = "alloc")))]
-mod key;
+pub mod key;
 #[cfg(feature = "alloc")]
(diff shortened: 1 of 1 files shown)


tarcieri:
@carl-wallace can you give an example of what this allows you to import which you couldn't import before?

As far as I can tell everything you need should be re-exported at the toplevel of the crate.

carl-wallace:
You are correct. Sorry about the noise. When I make the change below, the certval library builds and runs fine:
```
//use rsa::key::{PublicKey, RsaPublicKey};
//use rsa::pkcs8::DecodePublicKey;
use rsa::PublicKey;
use rsa::RsaPublicKey;
use rsa::pkcs8::FromPublicKey;
```

PR #143, opened 2022-02-02
Impl `zeroize` without using `zeroize_derive`

tarcieri:
The `zeroize_derive` crate has quite a few dependencies, and in particular `syn` is a fairly large one with not-insignificant compile times.

Together with dignifiedquire/num-bigint#35, this will remove the following dependencies from the dependency tree:

- `proc-macro2`
- `quote`
- `syn`
- `synstructure`
- `zeroize_derive`


tarcieri:
This should also take care of #142

PR #148, opened 2022-03-11
Replace rand dependency with rand_core

### LINKED ISSUE #115: CryptoRng requirement for RsaPrivateKey generate function
Is there a reason why the `new` and `new_with_exp` function of the `RsaPrivateKey` struct don't require the randomness source to implement `CryptoRng`? 

Requiring the `CryptoRng` impl would discourage using low entropy randomness sources at the type level as well as make it more consistent with other RustCrypto crates. The `ecdsa` crate, for example, requires the `SigningKey` type to be generated from a randomness source that implements `CryptoRng`

newpavlov:
`rand` is still a dev-dependency and used in code examples.

Also removes dependency on `rand_xorshift`, `base64`, and `hex`. Removes the `alloc`, which was effectively useless.

Closes #115

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -6,12 +6,9 @@
 //! Using PKCS1v15.
 //! ```
 //! use rsa::{PublicKey, RsaPrivateKey, RsaPublicKey, PaddingScheme};
-//! # /*
-//! use rand::rngs::OsRng;
-//! let mut rng = OsRng;
-//! # */
-//! # use rand::{SeedableRng, rngs::StdRng};
-//! # let mut rng = rand::rngs::StdRng::seed_from_u64(0);
+//!
+//! let mut rng = rand::thread_rng();
+//!
 //! let bits = 2048;
 //! let private_key = RsaPrivateKey::new(&mut rng, bits).expect("failed to generate a key");
 //! let public_key = RsaPublicKey::from(&private_key);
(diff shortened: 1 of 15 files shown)


src/lib.rs
```diff
@@ -6,12 +6,10 @@
 //! Using PKCS1v15.
 //! ```
 //! use rsa::{PublicKey, RsaPrivateKey, RsaPublicKey, PaddingScheme};
-//! # /*
-//! use rand::rngs::OsRng;
+//! use rand_core::OsRng;
```
tarcieri, line 9: I think it's helpful to re-export `rand_core`. 

It'd also be nice if there were a cargo feature to enable `getrandom` which toggles on `rand_core/getrandom`.

If there is a proper feature, we could also add helper methods which use `OsRng` automatically. Looking at the rest of the code, it looks like there are some places where that's already occurring and others where it's parameterized.

newpavlov: I am not sure whether we should use `OsRng` as a default RNG. `ThreadRng` looks like a better choice. We could add `rand` as an optional (but enabled by default) dependency and gate on it convenience methods which use `ThreadRng`.

tarcieri (approved):

newpavlov:
I will leave this PR open for several days to give @dignifiedquire time to review it.

PR #157, opened 2022-04-10
Fix rustdoc warnings and add rustdoc CI

tarcieri:
- Fixes some broken links to the `pkcs1`/`pkcs8` crate traits
- Fixes other rustdoc warnings
- Adds rustdoc CI job to ensure docs build without warnings


dignifiedquire (approved):
thanks

Issue #166, opened 2022-07-14
Denial of Service (DoS) via Large Public Key


missytake:
This issue was not discovered by me, but during and audit, see "L3" in https://delta.chat/assets/1907-otf-deltachat-rpgp-rustrsa-gb-reportv1.pdf:

"The RSA library allows operating upon large keys, which can consume a large amount of
computation time. An attacker who can force an application to encrypt with a million-byte RSA
public key can force the application into a Denial of Service (DoS) condition."

"The assessment team recommends exposing a higher-level API which performs additional
security checks. For instance, key sizes may be limited to 4096 bits by default but can be
overridden if necessary."

tarcieri:
Thanks for opening a tracking issue for this.

I agree that something like 4096 or 8192 would be a reasonable limit for a "checked" API.

My understanding is some applications (i.e. OpenPGP) need to support significantly larger keys. I'm not sure what the sanity limit should actually be there (IIRC @dignifiedquire mentioned up to 16384 iirc?)

So perhaps the default API could be changed to have a sanity limit, and an additional `*_unchecked` API could be added which doesn't support the sanity check to support exotic key sizes.

dignifiedquire:
So for gpg the default limit seems to be 8KiB, with a compile time option to go larger than that: https://wiki.gnupg.org/LargeKeys. 
I have not heard of any use case other than pgp for larger keys, so I think we could follow the same approach
- default: up to 4kib
- `medium-keys`: up to 8kib
- ´large-keys`: up to 16kib

tarcieri:
@dignifiedquire how about reducing that to just two options, one which allows keys <= 4096-bits, and "large keys" mode which supports up to 16384-bits?

tarcieri:
#170 restricts the maximum modulus size to 16384 bits. That's a stopgap, but not sufficient to close this issue.

To go further, we need to relegate keys that large to the previously proposed "large keys" mode, and set a lower bound which is used by default. I'm still thinking 4096-bits there.

tarcieri:
#176 now restricts the max public key size to 4096-bits, unless the special `RsaPublicKey::new_with_max_size` function is used to override the limit.

PR #167, opened 2022-07-24
Make `RsaPrivateKey::from_components` fallible

### LINKED ISSUE #163: `RsaPrivateKey::from_components` might panic
Some internals of the `num-bigint-dig` crate might panic under some conditions and `rsa` does not check for these. Therefore, users of the `rsa` crate experience panics.

I've stumbled upon this while implementing some jose stuff, so this examples uses a key in JsonWebKey format:

```rust
use base64ct::{Base64UrlUnpadded, Encoding};
use rsa::{BigUint, RsaPrivateKey};

fn main() -> Result<(), Box<dyn std::error::Error>> {
    let n = BigUint::from_bytes_be(&base64ct::Base64UrlUnpadded::decode_vec("w8XZRYEMzHbSMJXM6kSS2BGicmF5o9_cdlklKYXwey_HpWvI6XtRA9jq_SksGudyvhrLeyNr5P8mUoNPy9DiweXLoidv9kRbU33XLIvJkqbjIXKSwgk9RuQn-v8BoJj8e2uc7sH0eS3RVDtNZ1V1dRbyspC-e2zh9lgcgTRMUVM99pWijcrVD_Wt9hqYlFLM2HA_IoE-9gRrtqrRDwSOrISz_7pzRtJ3RFe3rVkh6MdZVUZ3bNa-D3A6acYmU74YFoVTyKlgSqNmuzFc9yOMM3z4amzLdvfeX_nUxWJdurK80M93vdTqw4-WIzlkkNkUOXq1LUjwZOvtke-t3y87JQ")?);

    let e = BigUint::from_bytes_be(&Base64UrlUnpadded::decode_vec("AQAB")?);

    let d = BigUint::from_bytes_be(&Base64UrlUnpadded::decode_vec("QThDVsVURzV6dpchKhZoOTU-wg45IN_uKTsvhzLI17EmOLS8vRPI_JgiSO6Tc-8RKcXxbfdx9VsPIEQArGzNbj0o5r9urEM_jYQJ0BxNrd6NIlJyE9RSJrDRpOuZVjBBRLioEl5pHImoCACtm7Q7qiNX_Sb9Xk76xD-8V0rd9eVJJBLXpoSV_b_Q80IB9YeZvE8VxzG9dzVjyowzn59SkvG-59G6SteOvLMb9i1in1dzXEKBzauYrExCOlT5t0skJenBVxgD-RNa9Bkjed_4QcNtO3x08F7j2ziqa8iNOtQWRThiArCd14iM2YhdHIZrzPeOYj9DmXIThOlacQ1NAQ")?);

    // original p value
    let p = BigUint::from_bytes_be(&Base64UrlUnpadded::decode_vec("11_iIv_gwowCkC32vALSm3Po9vfirnhX5jIyh2U2mzH9PvLx4OaUQ_O2WWyYRkpTNivSf0erFNjf9xz2efAhVrMX4Y11XPNpWAAsmuawA23Ptgjy6-PGSSHj8PJXXK6jTpoYe6uDtYIe509ODYdTeBU6KNg91T0EbNSV_A4K40E")?);

    let q = BigUint::from_bytes_be(&Base64UrlUnpadded::decode_vec("6LNts0tPvK6J2XgLaWI9rvvztJNzKTznmA5WPDpkApPpKp9U2lDucAyjAIFYJxYuOPN4kJ6tBeFGJAKEU1rv7hW5gXbbkVt6r_v23lV_LQ841pEnsmwvdUEgw3gE7iEviI6ZuiaSQQ-yUSoiKQc0rXeXhdmyVck_y5s1JHl_cuU")?);

    // malicious p
    // when set to zero, it panics and the backtrace points to https://docs.rs/num-bigint-dig/latest/src/num_bigint_dig/algorithms/sub.rs.html#75-79
    // when set to 1 it panics because of a divison by zero, probably because of https://docs.rs/rsa/latest/src/rsa/key.rs.html#344-345
    let p = BigUint::from_bytes_be(&[0x0]);

    // `precompute` in this method will panic
    let private_key = RsaPrivateKey::from_components(n, e, d, vec![p, q]);

    // it could be catched here (maybe, dont know tbh), but since it panics, there's
    // no way
    private_key.validate().unwrap();
    Ok(())
}
```

This can lead to denial of service attacks if untrusted keys are parsed and isn't mention in the docs at all.
I haven't tried malicious pem/der encoded keys, but it seems realistic that they suffer from the same problem.

I've also noticed that the `sign` method might loop if an incorrect parameter is used but that can probably be prevented by using `RsaPrivateKey::validate`.

Also, I really don't like the way `primes` are passed to `RsaPrivateKey::from_components` since passing an array with a length < 2 leads to panic because of out of bounds access and that isn't mention in the docs either.

tarcieri:
Adds an error case in the event the number of `primes` provides is fewer than 2, which prevents panics when invoking methods which expect primes to always be present at indices 0 and 1 (i.e. `p` and `q`)

Fixes #163


Cargo.toml
```diff
@@ -63,3 +63,6 @@ getrandom = ["rand_core/getrandom"]
 [package.metadata.docs.rs]
 features = ["std", "pem", "serde", "expose-internals"]
 rustdoc-args = ["--cfg", "docsrs"]
+
+[profile.dev]
+opt-level = 2
```
tarcieri, line 68: Threw this in there too. It significantly speeds up non-release tests.

tarcieri:
We could also add a check that the modulus is under some prescribed sanity limit (e.g. 16384 bits)

Erik1000:
> We could also add a check that the modulus is under some prescribed sanity limit (e.g. 16384 bits)

Yes, there should definitely be a sanity check. E.g. because of #166 since this would probably work with private keys too.

dignifiedquire (approved):
lgtm, thanks

dignifiedquire:
> We could also add a check that the modulus is under some prescribed sanity limit (e.g. 16384 bits)

Lets do that together with the compile time flags though, as that will be a (bigger) breaking change

PR #171, opened 2022-07-30
[WIP] Bound `RsaPublicKey::new` on 4096-bits; add `::new_large`

### LINKED ISSUE #166: Denial of Service (DoS) via Large Public Key

tarcieri:
*NOTE: depends on #170*

This commit fixes #166 by enforcing a 4096-bit upper limit by default, which prevents potential DoS by using maliciously large RSA keys.

The PKCS#1/PKCS#8 parsers use this API, limiting the size of keys parsed from these formats to 4096-bits.

An `RsaPrivateKey::new_large` constructor has been added which enforces the 16384-bit limit added in #170. This can be used for unusual applications that need to support larger keys.

`RsaPrivateKey::from_components` uses the `::new_large` constructor, so private keys follow the 16384-bit limit only.

The `RsaPrivateKey::MAX_SIZE` and `RsaPrivateKey::MAX_SIZE_LARGE` inherent constants specify the maximum allowed sizes.


src/key.rs
```diff
@@ -228,8 +225,39 @@ impl PublicKey for RsaPublicKey {
 }
 
 impl RsaPublicKey {
-    /// Create a new key from its components.
+    /// Minimum value of the public exponent `e`.
+    pub const MIN_PUB_EXPONENT: u64 = 2;
+
+    /// Maximum value of the public exponent `e`.
+    pub const MAX_PUB_EXPONENT: u64 = (1 << 33) - 1;
```
tarcieri, line 232: Feels a bit weird to make these `pub` as `u64`. I think it'd be nice to have access to them, but there is a mismatch between `u64` here and `BigUint` when passed as a parameter.

Perhaps we should make `e: u64` as well?

aloucks:
Having two APIs that are essentially the same thing feels a little clunky. Why not add a `Limits` struct and a `Default` as `4096`?

Something like:

```rust
pub struct Limits {
  /// The maximum key size in bytes
  pub max_key_size: usize,
}

impl Default for Limits {
  fn default() -> Self {
    Limits { max_key_size: 4096 }
  }
}

pub fn new(n: BigUint, e: BigUint, limits: Limits) -> Result<Self> {
  // ...
}
```

tarcieri:
That needlessly complicates the common-case API.

However, we could have two methods, where one is composed in terms of the other, e.g. `new` vs `new_with_max_key_size`

tarcieri:
Closing in favor of #176

Issue #172, opened 2022-07-31
PSS signature verification should not require rng


lumag:
Currently creation of PSS PaddingSchema object requires the RngCore, however if the software is only verifying the signature, it should not require the random number generator. Please consider changing the rng to Option<> or adding `new_verify_pss` / `new_verify_pss_with_salt`.

tarcieri:
I left some notes on #172, but in general I think the signature code is really due for a refactor.

I'd need to write up a more concrete plan for what I have in mind though. It would be nice to actually impl the traits from the [`signature`](https://docs.rs/signature) crate.

lumag:
I tried looking onto the `signature`-based implementation, but stopped after seeing no good way to implement `RsaSignature`. Unlike ECDSA, where the signature is bound by the size of the curve/field, for the RSA there is no such limitation. Or I can add the artificial limitation of not supporting anything above RSA16384. Does that sound good?

tarcieri:
I can open a new issue and sketch out a more complete design

PR #173, opened 2022-07-31
Implement verify-only PSS PaddingScheme

lumag:
A fix for #172 , implementation of verify-only PSS padding schema

### DIFF
--- a/src/key.rs
+++ b/src/key.rs
@@ -512,18 +512,24 @@ impl RsaPrivateKey {
             PaddingScheme::PKCS1v15Sign { ref hash } => {
                 pkcs1v15::sign::<DummyRng, _>(None, self, hash.as_ref(), digest_in)
             }
+            _ => Err(Error::InvalidPaddingScheme),
+        }
+    }
+
+    /// Sign the given digest using the provided rng
+    ///
+    /// Use `rng` for signature process.
+    pub fn sign_with_rng<R: RngCore + CryptoRng>(
+        &self,
+        rng: &mut R,
+        padding: PaddingScheme,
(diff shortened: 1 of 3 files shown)


tarcieri:
Instead of introducing a `DummyRng`, it seems to me like it would make more sense to remove `salt_rng` from `PaddingScheme::PSS`.

Its only use is to pass it to `pss::sign`, and in that regard it can be passed as an explicit parameter rather than embedding it into the padding scheme. An RNG is already an explicit parameter to `sign_blinded` (in which case it receives two RNGs. I'm not sure why one would want to use two RNGs?)

Really it'd be nice to implement the traits from the `signature` crate, where the "sign" vs "sign with RNG" cases are concretely handled as [`Signer`](https://docs.rs/signature/latest/signature/trait.Signer.html) versus [`RandomizedSigner`](https://docs.rs/signature/latest/signature/trait.RandomizedSigner.html), but that's a lot more work.

lumag:
@tarcieri done

lumag:
gracious ping

lumag:
Refreshed to remove conflict.

tarcieri (approved):
Looks a lot better!

dignifiedquire (approved):
lgtm thanks

Kriskras99:
@tarcieri could this be included in the upcoming 0.7.0 release? It would save me a dependency and make it clearer what is going on in my code.

tarcieri:
Yep, that's the plan

Kriskras99:
> Yep, that's the plan

Great! I look forward to testing 0.7.0-pre2😁

PR #174, opened 2022-07-31
Implement Signer/Verifier/Signature interfaces for the RSA signatures

lumag:
Refactor the `rsa` crate to use the API defined by the `signature` crate.
This adds `pss` and `pkcs1v15` modules, each of them providing `Signature`, `Verifier` and `Signer`/`RandomizedSigner` implementations.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -23,6 +23,8 @@ subtle = { version = "2.1.1", default-features = false }
 digest = { version = "0.10.0", default-features = false, features = ["alloc"] }
 pkcs1 = { version = "0.4", default-features = false, features = ["pkcs8", "alloc"] }
 pkcs8 = { version = "0.9", default-features = false, features = ["alloc"] }
+#To keep the rand_core versions properly pinnen, specify exact version
+signature = { version = ">=1.5, <1.7", default-features = false , features = ["rand-preview"] }
 zeroize = { version = "1", features = ["alloc"] }
 
 # Temporary workaround until https://github.com/dignifiedquire/num-bigint/pull/42 lands
@@ -55,7 +57,7 @@ default = ["std", "pem"]
 nightly = ["num-bigint/nightly"]
 serde = ["num-bigint/serde", "serde_crate"]
 expose-internals = []
-std = ["digest/std", "pkcs1/std", "pkcs8/std", "rand_core/std"]
+std = ["digest/std", "pkcs1/std", "pkcs8/std", "rand_core/std", "signature/std"]
 pem = ["pkcs1/pem", "pkcs8/pem"]
 pkcs5 = ["pkcs8/encryption"]
 getrandom = ["rand_core/getrandom"]
--- a/src/pkcs1v15.rs
+++ b/src/pkcs1v15.rs
@@ -1,18 +1,89 @@
 use alloc::vec;
 use alloc::vec::Vec;
+use core::fmt::{Debug, Display, Formatter, LowerHex, UpperHex};
 use rand_core::{CryptoRng, RngCore};
+use signature::{RandomizedSigner, Signature as SignSignature, Signer, Verifier};
 use subtle::{Choice, ConditionallySelectable, ConstantTimeEq};
 use zeroize::Zeroizing;
 
+use crate::dummy_rng::DummyRng;
 use crate::errors::{Error, Result};
 use crate::hash::Hash;
 use crate::key::{self, PrivateKey, PublicKey};
+use crate::{RsaPrivateKey, RsaPublicKey};
+
+#[derive(Clone)]
+pub struct Signature {
+    bytes: Vec<u8>,
+}
+
+impl signature::Signature for Signature {
+    fn from_bytes(bytes: &[u8]) -> signature::Result<Self> {
+        Ok(Signature {
+            bytes: bytes.into(),
+        })
+    }
+
+    fn as_bytes(&self) -> &[u8] {
+        &self.bytes.as_slice()
+    }
+}
+
+impl From<Vec<u8>> for Signature {
+    fn from(bytes: Vec<u8>) -> Self {
+        Self { bytes }
+    }
+}
+
+impl PartialEq for Signature {
+    fn eq(&self, other: &Self) -> bool {
+        self.as_bytes() == other.as_bytes()
+    }
+}
+
+impl Eq for Signature {}
+
+impl Debug for Signature {
+    fn fmt(&self, fmt: &mut Formatter<'_>) -> core::result::Result<(), core::fmt::Error> {
+        fmt.debug_list().entries(self.as_bytes().iter()).finish()
+    }
+}
+
+impl AsRef<[u8]> for Signature {
+    fn as_ref(&self) -> &[u8] {
+        &self.as_bytes()
+    }
+}
+
+impl LowerHex for Signature {
+    fn fmt(&self, f: &mut Formatter<'_>) -> core::fmt::Result {
+        for byte in self.as_bytes() {
+            write!(f, "{:02x}", byte)?;
+        }
+        Ok(())
+    }
+}
+
+impl UpperHex for Signature {
+    fn fmt(&self, f: &mut Formatter<'_>) -> core::fmt::Result {
+        for byte in self.as_bytes() {
+            write!(f, "{:02X}", byte)?;
+        }
+        Ok(())
+    }
+}
+
+impl Display for Signature {
+    fn fmt(&self, f: &mut Formatter<'_>) -> core::fmt::Result {
+        write!(f, "{:X}", self)
+    }
+}
 
 // Encrypts the given message with RSA and the padding
 // scheme from PKCS#1 v1.5.  The message must be no longer than the
 // length of the public modulus minus 11 bytes.
 #[inline]
-pub fn encrypt<R: RngCore + CryptoRng, PK: PublicKey>(
+pub(crate) fn encrypt<R: RngCore + CryptoRng, PK: PublicKey>(
     rng: &mut R,
     pub_key: &PK,
     msg: &[u8],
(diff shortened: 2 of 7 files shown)


tarcieri:
As far as the structure goes, I'd suggest doing something similar to the https://github.com/rustcrypto/elliptic-curves crates, which define the following types inside of modules named after the signature algorithm:

- `Signature`
- `SigningKey`
- `VerifyingKey`

See the following modules as examples:

- https://docs.rs/k256/latest/k256/ecdsa/index.html
- https://docs.rs/k256/latest/k256/schnorr/index.html

...then `impl From<RsaPrivateKey> for SigningKey` and `impl From<&RsaPrivateKey> for SigningKey` (and so on)

For PSS, parameters like the MGF digest can be generic parameters of the key and/or signature, e.g.

```rust
pub struct Signature<Mgf: Digest> {
    data: Vec<u8>,
    mgf: PhantomData<Mgf>
}
````

**Comment by lumag** [comment]
> ...then `impl From<RsaPrivateKey> for SigningKey` and `impl From<&RsaPrivateKey> for SigningKey` (and so on)
> 
> For PSS, parameters like the MGF digest can be generic parameters, e.g.
> 
> ```rust
> pub struct SigningKey<Mgf: Digest> {
>     inner: RsaPrivateKey,
>     mgf: PhantomData<Mgf>,
>     [...]
> }
> ```

What about the `salt_len`? Is it fine to set it via `fn pss::SigningKey::set_salt_len(&self, Option<usize>);` ?

**Comment by tarcieri** [comment]
@lumag yep!

**Comment by lumag** [comment]
@tarcieri I have mostly finished the implementation of Signer/Verifier implementations. However I now have an issue with the PSS Signer. The trait passes self as non-mutable object, while the signing function uses salt rng as mutable.

**Comment by tarcieri** [comment]
>  The trait passes self as non-mutable object, while the signing function uses salt rng as mutable.

`RandomizedSigner`/`RandomizedDigestSigner` accept the RNG as an explicit parameter so you don't have to shove it in `self`.

**Comment by lumag** [comment]
@tarcieri pushed next iteration following your comments.

File: src/pkcs1v15.rs [hunk 1]
```diff
@@ -214,6 +219,162 @@ fn non_zero_random_bytes<R: RngCore + CryptoRng>(rng: &mut R, data: &mut [u8]) {
     }
 }
 
+pub struct RsaPkcs1v15SigningKey {
```
tarcieri, line 222: These names seem to run afoul of the "no module prefix" convention:

https://rust-lang.github.io/rfcs/0356-no-module-prefixes.html

i.e. the full name is:

`rsa::pkcs1v15::RsaPkcs1v15SigningKey`

Instead it could just be:

`rsa::pkcs1v15::SigningKey`

This allows someone to import `rsa::pkcs1v15` and get access to:

- `pkcs1v15::SigningKey`
- `pkcs1v15::VerifyingKey`
- `pkcs1v15::Signature`

...rather than having to import three types that all start with the module name.

src/pkcs1v15.rs
```diff
@@ -214,6 +219,162 @@ fn non_zero_random_bytes<R: RngCore + CryptoRng>(rng: &mut R, data: &mut [u8]) {
...
+                key.primes().clone().to_vec(),
+            )
+            .unwrap(),
+            hash: None,
+        }
+    }
+}
+
+impl From<(RsaPrivateKey, Hash)> for RsaPkcs1v15SigningKey {
+    fn from(data: (RsaPrivateKey, Hash)) -> Self {
```
tarcieri, line 262: This feels like it should probably just be:

```rust
impl SigningKey {
    pub fn new(key: RsaPrivateKey, hash: Hash) -> Self {
```

src/pss.rs
```diff
@@ -235,15 +264,177 @@ fn emsa_pss_verify(
     }
 }
 
+pub struct RsaPSSSigningKey {
```
tarcieri, line 267: Same comment here re: `pss::SigningKey` and the "no module prefix" convention

src/signature.rs
```diff
@@ -0,0 +1,70 @@
+use alloc::vec::Vec;
+use core::fmt::{Debug, Display, Formatter, LowerHex, UpperHex};
+use signature::Signature as SignSignature;
+
+#[derive(Clone)]
+pub struct Signature {
+    bytes: Vec<u8>,
+}
+
+impl SignSignature for Signature {
```
tarcieri, line 10:
```suggestion
impl signature::Signature for Signature {
```

src/pkcs1v15.rs
```diff
@@ -1,12 +1,17 @@
...
 use subtle::{Choice, ConditionallySelectable, ConstantTimeEq};
 use zeroize::Zeroizing;
 
+use crate::dummy_rng::DummyRng;
 use crate::errors::{Error, Result};
 use crate::hash::Hash;
 use crate::key::{self, PrivateKey, PublicKey};
+use crate::{PublicKeyParts, RsaPrivateKey, RsaPublicKey};
+
+pub use crate::Signature;
```
tarcieri, line 14: I'd suggest giving `pkcs1v15` and `pss` different `Signature` newtypes in order to make them type safe so they don't get confused for one another

tarcieri:
Looking better!

src/pkcs1v15.rs
lumag: Ack. Initially I had two different types, which had 100% of common code. I'll change this back.

src/pkcs1v15.rs
lumag: Heh. At some point you wrote about using `From`, so I followed that suggestion blindly. I'll change this to use `::new()`.

lumag:
@tarcieri done

Cargo.toml
```diff
@@ -23,6 +23,7 @@ subtle = { version = "2.1.1", default-features = false }
 digest = { version = "0.10.0", default-features = false, features = ["alloc"] }
 pkcs1 = { version = "0.4", default-features = false, features = ["pkcs8", "alloc"] }
 pkcs8 = { version = "0.9", default-features = false, features = ["alloc"] }
+signature = { version = "1.5", default-features = false , features = ["rand-preview"] }
```
tarcieri, line 26: To keep the `rand_core` versions properly pinned you need to do:

```suggestion
signature = { version = ">=1.5, <1.7", default-features = false , features = ["rand-preview"] }
```

tarcieri:
Looks mostly good to me now.

@lumag can you update the PR description?

Cargo.toml
lumag: Done

lumag:
@tarcieri updated the description

tarcieri (approved):

vccggorski:
Is it necessary to couple `Signature` type with std/alloc Vec? Maybe having `Signature<T: AsRef<[u8]>> { bytes: T }` would give more freedom, especially when reusing a type to implement a signature verifier on some embedded allocless platform. I'm not sure how feasible linking against RSA crate without alloc really is anyway, maybe `pss::Signature` or any other `<specific>::Signature` could be kept in some other lightweight crate that could provide some convenience aliases for specific platforms that would use `Vec` as is?

```rust
pub struct Signature<T: AsRef<[u8]>> {
    bytes: T
}
pub mod std_alloc {
    pub type Signature = super::Signature<alloc::vec::Vec<u8>>;
}
// proper re-exports depending on `std` feature?

vccggorski:
I realized that `Signature` has a hard requirement to be constructible from arbitrary slices (`signature::Signature::from_bytes`). That might make this exercise a little bit harder.

tarcieri:
@vccggorski the `rsa` crate has a hard dependency on liballoc for the `BigUint` representation (in `num-bigint-dig`). So the crate lacks heapless support in general.

There's been some discussion of heapless support using stack-allocated bigints, e.g. via the `crypto-bigint` crate.

> I realized that Signature has a hard requirement to be constructible from arbitrary slices

The constructor is fallible, so that is not a problem at all. The `ecdsa` and `ed25519` crates both provide stack-allocated types which impl `Signature`.

lumag:
Initially I had a separate alloc-less `Signature` implementation using on-stack array. However after noticing that the rest of the code (both pss and pkcs1) heavily depend on alloc, I dropped this implementation. If at some point the RSA crates moves towards alloc-less implementation of bigint, then both `Signature` and dynamic digests code can be changed.

vccggorski:
Also by reading the code I got a little bit confused because my understanding of a difference between `DigestVerifier/Signer` and `Verifier/Signer` is that former operates on a hash of a message whereas latter needs to generate a hash/digest first. This is why you could derive `Verifier` out of `DigestVerifier` if your `Signature` is a `PreHashedSignature` and knows what is its associated hashing method. Here, `pss::VerifyingKey` implements _just_ `Verifier` and assumes that a `message` is a hash. Sorry if I'm confusing things, I'm still trying to wrap my head around these concepts.

tarcieri:
The `Digest*` traits are more flexible in that they allow you to pass alternative hash functions, or alternative implementations of hash functions (e.g. wrapper for a hardware accelerator)

lumag:
I did not see a particular value of implementing the `DigestSigner` traits, especially given their description. However I can implement them if required.

tarcieri:
@lumag the nice thing would be using a hardware accelerator for hashing, but given this crate isn't particularly embedded-friendly to begin with I think leaving that out would be fine for an MVP

lumag:
@tarcieri ok, I will add them to my queue to take a look after sorting out the `From<&SK>` story.

tarcieri:
I'm going to go ahead and merge this.

@lumag feel free to follow up with `Digest*` impls if you'd like, although it seems like there are additional features you're working on which are more important.

lumag:
@sandhose @tarcieri Well, I followed the previous design of signing the pre-hashed messages, since this looks to me the way how the RSA signatures usually work. However I'm fine with changing that to use raw messages and hashing them during the sign procedure. The only question is about the raw PKCS1 v1.5 signatures (which do not have the ASN.1 wrapping).

tarcieri:
@sandhose good catch! The `Signer` and `Verifier` traits should prehash the message.

@lumag it would probably be good to retain inherent methods which operate on the bytes of a message digest, then implement traits like `DigestSigner` and `Signer` in terms of the inherent methods.

Generally working directly with raw digests is an antipattern. It's much less of a problem with RSA than it is with e.g. ECDSA or Schnorr though, where it can lead to existential forgeries.

lumag:
@sandhose https://github.com/RustCrypto/RSA/pull/178

PR #175, opened 2022-08-08
Fix bug on pkcs1.rs

jeppeter:
encode_rsa2048_priv_pem
   		encode_rsa4096_priv_pem
   		encode_rsa2048_pub_pem
   		encode_rsa4096_pub_pem
   pkcs8.rs
   		encode_rsa2048_priv_pem
   		encode_rsa2048_pub_pem
   when running in windows-gnu adding the '\r' in string

### DIFF
--- a/tests/pkcs1.rs
+++ b/tests/pkcs1.rs
@@ -169,7 +169,8 @@ fn decode_rsa4096_pub_pem() {
 fn encode_rsa2048_priv_pem() {
     let key = RsaPrivateKey::from_pkcs1_pem(RSA_2048_PRIV_PEM).unwrap();
     let pem = key.to_pkcs1_pem(Default::default()).unwrap();
-    assert_eq!(&*pem, RSA_2048_PRIV_PEM)
+    let c = pem.replace("\r","");
+    assert_eq!(c, RSA_2048_PRIV_PEM)
 }
 
 #[test]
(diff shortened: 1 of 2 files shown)


tests/pkcs1.rs
```diff
@@ -169,29 +169,33 @@ fn decode_rsa4096_pub_pem() {
 fn encode_rsa2048_priv_pem() {
     let key = RsaPrivateKey::from_pkcs1_pem(RSA_2048_PRIV_PEM).unwrap();
     let pem = key.to_pkcs1_pem(Default::default()).unwrap();
```
tarcieri, line 171: It's inserting the `\r` because `Default::default()` is actually `LineEnding::default()`, which uses OS-specific line endings.

To make the test always use `\n` line endings regardless of OS, you can use:

```suggestion
    let pem = key.to_pkcs1_pem(LineEnding::LF).unwrap();
```

aloucks:
You might want to add a `.gitattributes` that tells git that `.der` files use `eol=LF` so that it doesn't ever try to rewrite the line endings.

tarcieri:
`.der` files are binary. They don't have line endings

aloucks:
> `.der` files are binary. They don't have line endings.
> 
> Do you mean `.pem`?

Yes

PR #176, opened 2022-08-08
Add `RsaPublicKey::new_with_max_size`; 4096-bit default max

### LINKED ISSUE #166: Denial of Service (DoS) via Large Public Key

tarcieri:
This constructor accepts a configurable maximum key size which can be used in applications that need to deal with unusually large RSA public keys, such as OpenPGP.

With the ability to configure an upper limit, this makes it possible to lower the default maximum key size to 4096-bits, which is a reasonable upper limit for most applications.

Fixes #166

### DIFF
--- a/src/errors.rs
+++ b/src/errors.rs
@@ -1,7 +1,7 @@
 pub type Result<T> = core::result::Result<T, Error>;
 
 /// Error types
-#[derive(Debug)]
+#[derive(Debug, Eq, PartialEq)]
 #[non_exhaustive]
 pub enum Error {
     InvalidPaddingScheme,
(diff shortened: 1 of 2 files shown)


dignifiedquire (commented):
lgtm, can you add a test for the new defaults?

tarcieri:
@dignifiedquire added

dignifiedquire (approved):

PR #179, opened 2022-08-30
Update RSA signature traits implementations

lumag:
- [X] Change `Signer`/`Verifier` implementations to accept raw messages rather than pre-hashed data
- [X] Implement `DigestSigner`/`Verifier` traits
- [x] Get rid of `DynDigest` bounds for the RSASSA-PSS implmentation
- [x] Sort out specifying Hash to the `pkcs1v15::SigningKey<D>::new_with_hash`, which duplicates information passed in `<D>`
- [ ]  ~~Use passed `digest` object for PSS calculations rather than newly created `Digest`?~~

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -30,6 +30,11 @@ zeroize = { version = "1", features = ["alloc"] }
 # Temporary workaround until https://github.com/dignifiedquire/num-bigint/pull/42 lands
 smallvec = { version = "1.6.1", default-features = false }
 
+# Temporary until the link from Digest to OID is moved to corresponding crates
+sha1 = { version = "0.10.1", default-features = false, optional = true }
+sha2 = { version = "0.10.2", default-features = false, optional = true }
+sha3 = { version = "0.10.1", default-features = false, optional = true }
+
 [dependencies.serde_crate]
 package = "serde"
 optional = true
@@ -53,7 +58,7 @@ sha3 = { version = "0.10.1", default-features = false }
 name = "key"
 
 [features]
-default = ["std", "pem"]
+default = ["std", "pem", "sha2"]
 nightly = ["num-bigint/nightly"]
 serde = ["num-bigint/serde", "serde_crate"]
 expose-internals = []
@@ -63,7 +68,7 @@ pkcs5 = ["pkcs8/encryption"]
 getrandom = ["rand_core/getrandom"]
 
 [package.metadata.docs.rs]
-features = ["std", "pem", "serde", "expose-internals"]
+features = ["std", "pem", "serde", "expose-internals", "sha1", "sha2", "sha3"]
 rustdoc-args = ["--cfg", "docsrs"]
 
 [profile.dev]
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -24,7 +24,7 @@ digest = { version = "0.10.0", default-features = false, features = ["alloc"] }
 pkcs1 = { version = "0.4", default-features = false, features = ["pkcs8", "alloc"] }
 pkcs8 = { version = "0.9", default-features = false, features = ["alloc"] }
 #To keep the rand_core versions properly pinnen, specify exact version
-signature = { version = ">=1.5, <1.7", default-features = false , features = ["rand-preview"] }
+signature = { version = ">=1.5, <1.7", default-features = false , features = ["digest-preview", "rand-preview"] }
 zeroize = { version = "1", features = ["alloc"] }
 
 # Temporary workaround until https://github.com/dignifiedquire/num-bigint/pull/42 lands
--- a/src/hash.rs
+++ b/src/hash.rs
@@ -83,3 +83,55 @@ impl Hash {
         }
     }
 }
+
+/* FIXME: This trait should be refactored into per-digest implementations returning OID */
+pub trait AssociatedHash {
+    const HASH: Hash;
+}
+
+#[cfg(feature = "sha1")]
+impl AssociatedHash for sha1::Sha1 {
+    const HASH: Hash = Hash::SHA1;
+}
+
+#[cfg(feature = "sha2")]
+impl AssociatedHash for sha2::Sha224 {
+    const HASH: Hash = Hash::SHA2_224;
+}
+
+#[cfg(feature = "sha2")]
+impl AssociatedHash for sha2::Sha256 {
+    const HASH: Hash = Hash::SHA2_256;
+}
+
+#[cfg(feature = "sha2")]
+impl AssociatedHash for sha2::Sha384 {
+    const HASH: Hash = Hash::SHA2_384;
+}
+
+#[cfg(feature = "sha2")]
+impl AssociatedHash for sha2::Sha512 {
+    const HASH: Hash = Hash::SHA2_512;
+}
+
+/*
+#[cfg(feature = "sha3")]
+impl AssociatedHash for sha3::Sha3_224 {
+    const HASH: Hash = Hash::SHA3_224;
+}
+*/
+
+#[cfg(feature = "sha3")]
+impl AssociatedHash for sha3::Sha3_256 {
+    const HASH: Hash = Hash::SHA3_256;
+}
+
+#[cfg(feature = "sha3")]
+impl AssociatedHash for sha3::Sha3_384 {
+    const HASH: Hash = Hash::SHA3_384;
+}
+
+#[cfg(feature = "sha3")]
+impl AssociatedHash for sha3::Sha3_512 {
+    const HASH: Hash = Hash::SHA3_512;
+}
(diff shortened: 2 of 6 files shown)


lumag:
RFC:

- `DynDigest` in PSS implementations. I did not want to touch internal implementation (yet), which uses `DynDigest` internally. The easiest way to drop the bound is to implement internal dynamic `DynDigest` wrapper around passed `D: Digest` type. More proper would be to change/duplicate `emsa_pss_encode` and `emsa_pss_verify` into generic functions. Which path should I implement?

- `Hash` duplicates D type argument. I have thought about adding new trait and adding a type boundary to the Pkcs1v15 type argument. It should probably go to the `pkcs1` crate. However implementing it would bring in at least sha1/sha2/sha3 dependencies.
 
  ```Rust
  pub trait DigestWithOID: Digest {
      fn get_oid() -> &'static [u8];
  }
  ```

  Another approach might be to change `new_with_hash` argument to accept OID/prefix directly. But then we are just changing the form of information duplication instead of removing it.

- Regarding using passed `digest`. It is closely tied to the first point. Plus I was not sure if it is okay to reuse the external struct.

sandhose:
> DynDigest in PSS implementations. I did not want to touch internal implementation (yet), which uses DynDigest internally. The easiest way to drop the bound is to implement internal dynamic DynDigest wrapper around passed D: Digest type. More proper would be to change/duplicate emsa_pss_encode and emsa_pss_verify into generic functions. Which path should I implement?

Shouldn't `dyn DynDigest` implement `Digest`? If it's the case, I think you could have `emsa_pss_encode` be `emsa_pss_encode<D: Digest>(..., digest: &mut D)`, and still use it with a `&dyn DynDigest`/`Box<dyn DynDigest>`

> Hash duplicates D type argument. I have thought about adding new trait and adding a type boundary to the Pkcs1v15 type argument. It should probably go to the pkcs1 crate. However implementing it would bring in at least sha1/sha2/sha3 dependencies.
> 
> ```rust
> pub trait DigestWithOID: Digest {
>     fn get_oid() -> &'static [u8];
> }
> ```
> 
> Another approach might be to change new_with_hash argument to accept OID/prefix directly. But then we are just changing the form of information duplication instead of removing it.

Looks like there is `const_oid::AssociatedOid` trait for that: https://docs.rs/const-oid/latest/const_oid/trait.AssociatedOid.html
Not 100% sure it's the right thing, but it's implemented for elliptic curves. If it is, it's then a matter of implementing them in the various digest crates

lumag:
> > DynDigest in PSS implementations. I did not want to touch internal implementation (yet), which uses DynDigest internally. The easiest way to drop the bound is to implement internal dynamic DynDigest wrapper around passed D: Digest type. More proper would be to change/duplicate emsa_pss_encode and emsa_pss_verify into generic functions. Which path should I implement?
> 
> Shouldn't `dyn DynDigest` implement `Digest`? If it's the case, I think you could have `emsa_pss_encode` be `emsa_pss_encode<D: Digest>(..., digest: &mut D)`, and still use it with a `&dyn DynDigest`/`Box<dyn DynDigest>`

No, there are minor differences there, so `Digest` and `DynDigest` are not directly interchangeable. However. Let me check if it would be possible to wrap it using `Fn`.

> > Hash duplicates D type argument. I have thought about adding new trait and adding a type boundary to the Pkcs1v15 type argument. It should probably go to the pkcs1 crate. However implementing it would bring in at least sha1/sha2/sha3 dependencies.
> > ```rust
> > pub trait DigestWithOID: Digest {
> >     fn get_oid() -> &'static [u8];
> > }
> > ```
> >   
> > Another approach might be to change new_with_hash argument to accept OID/prefix directly. But then we are just changing the form of information duplication instead of removing it.
> 
> Looks like there is `const_oid::AssociatedOid` trait for that: https://docs.rs/const-oid/latest/const_oid/trait.AssociatedOid.html Not 100% sure it's the right thing, but it's implemented for elliptic curves. If it is, it's then a matter of implementing them in the various digest crates

This looks like what I was looking for. Let's look it using it (and implementing it in various digest crates) would be fine with @tarcieri .

newpavlov:
>Shouldn't dyn DynDigest implement Digest? 

It's not possible. The main difference is that `Digest` returns hash results as arrays, which makes it non-object-safe, while `DynDigest` works around it by returning hash results as boxed slices. Imagine code which tries to simultaneously work with both SHA-1 and SHA-256.

It's an architecture decision: you should decide whether you rely on dynamic dispatch (i.e. `DynDigest`) or on monomorphization (i.e. `Digest`). I don't think it makes much sense to mix them in the same code.

lumag:
@tarcieri @dignifiedquire your opinion/comments?

Cargo.toml
```diff
@@ -58,6 +58,7 @@ nightly = ["num-bigint/nightly"]
 serde = ["num-bigint/serde", "serde_crate"]
 expose-internals = []
 std = ["digest/std", "pkcs1/std", "pkcs8/std", "rand_core/std", "signature/std"]
+digest-preview = ["signature/digest-preview" ]
```
tarcieri, line 61: The `rsa` crate already has a hard dependency on `digest`, so you can just enable this unilaterally:

```toml
signature = { version = ">=1.5, <1.7", default-features = false , features = ["digest-preview", "rand-preview"] }
```

src/lib.rs
```diff
@@ -58,17 +58,16 @@
 //!
 //! let bits = 2048;
 //! let private_key = RsaPrivateKey::new(&mut rng, bits).expect("failed to generate a key");
-//! let signing_key = SigningKey::new_with_hash(private_key, Hash::SHA2_256);
-//! let verifying_key: VerifyingKey = (&signing_key).into();
+//! let signing_key = SigningKey::<Sha256>::new_with_hash(private_key, Hash::SHA2_256);
```
tarcieri, line 61:
```rust
SigningKey::<Sha256>::new_with_hash(private_key, Hash::SHA2_256);
```

...still repeats `Sha256` and `Hash::SHA2_256`.

lumag: This is a bullet 3 in the RFC in the first comment. @sandhose suggested implementing `const_oid::AssociatedOID for the `Digest` implementations instead of having a separate `Hash` class. Is that ok with you?

tarcieri:
@lumag using a trait to provide the linkage between a generic `D: Digest` and the current usages of `Hash` is a good idea.

Will `AssociatedOid` actually work for this use case though? Aren't the OIDs that matter here RSA-specific (e.g. MGF)?

If you use `AssociatedOid`, the relevant OIDs will need to be added to types defined in the `sha2` crate. And if we do that, they need to be the generic OIDs that aren't predicated around RSA-specific use cases, e.g. [`id-sha256` (2.16.840.1.101.3.4.2.1)](http://oid-info.com/get/2.16.840.1.101.3.4.2.1)

lumag:
@tarcieri no, the OIDs are generic enough. Consider e.g. `Hash::SHA2_256`:
```Rust
Hash::SHA2_256 => &[
                0x30, 0x31, 0x30, 0x0d, 0x06, 0x09, 0x60, 0x86, 0x48, 0x01, 0x65, 0x03, 0x04, 0x02,
                0x01, 0x05, 0x00, 0x04, 0x20,
            ],
```

This translates to:

```
  0  49: SEQUENCE {
  2  13:   SEQUENCE {
  4   9:     OBJECT IDENTIFIER sha-256 (2 16 840 1 101 3 4 2 1)
 15   0:     NULL
       :     }
 17  32:   OCTET STRING
```

lumag:
I'll do that, I'll need to test it anyway.

tarcieri:
@lumag maybe for now you can add a trait with an associated constant of the existing `Hash` type?

Using OIDs seems interesting but yes that's a complication. Semi-related issue: https://github.com/RustCrypto/traits/issues/1069

lumag:
This will add dependencies from the rsa crate on the sha1/sha2/sha3 crates. Is it fine for now?

tarcieri:
@lumag maybe optional dependencies with `sha2` enabled by default?

lumag:
@tarcieri done

src/hash.rs
```diff
@@ -83,3 +83,73 @@ impl Hash {
         }
     }
 }
+
+/* FIXME: This trait should be refactored into per-digest implementations returning OID */
+pub trait AssociatedHash {
+    fn get_hash() -> Hash;
```
tarcieri, line 89:
```suggestion
    const HASH: Hash;
```

lumag: Done

tarcieri (approved):

tarcieri:
Going to go ahead and merge this as it's hopefully uncontroversial additive changes to the new signature features.

@lumag seems we have a path forward on replacing `Hash` with OIDs if you'd like to attempt that as a followup: https://github.com/RustCrypto/traits/pull/1098

lumag:
@tarcieri thanks!

PR #182, opened 2022-09-07
Allow dereferencing Signature types

lumag:
Implement Deref<Target = [u8]> for the Signature types to allow automatically dereferencing Signature as byte slices.


tarcieri (approved):
Seems like an OK "smart pointer" case.

Issue #151, opened 2022-03-13
Add changelog


newpavlov:
Ideally it would be nice to add it before v0.6 release.

PR #183, opened 2022-09-09
pkcs1v15: use AssociatedOID for getting the RSA prefix

lumag:
Drop internal implementation of AssociatedHash and use AssociatedOID
trait to get the OID corresponding to the Digest and to format the ASN.1
prefix.


lumag:
This depends on https://github.com/RustCrypto/traits/pull/1098 and https://github.com/RustCrypto/hashes/pull/405 to be merged first.

Cargo.toml
```diff
@@ -50,15 +45,15 @@ rand_xorshift = "0.3"
 rand_chacha = "0.3"
 rand = "0.8"
 rand_core = { version = "0.6", default-features = false }
-sha1 = { version = "0.10.1", default-features = false }
-sha2 = { version = "0.10.2", default-features = false }
+sha1 = { version = "0.10.5", default-features = false, features = ["oid"] }
+sha2 = { version = "0.10.6", default-features = false, features = ["oid"] }
 sha3 = { version = "0.10.1", default-features = false }
```
newpavlov, line 50: Maybe it's worth to add tests for SHA-3 OIDs as well?

src/hash.rs
```diff
@@ -1,137 +0,0 @@
...
-                0x02, 0x05, 0x00, 0x04, 0x30,
-            ],
-
-            Hash::SHA2_512 => &[
-                0x30, 0x51, 0x30, 0x0d, 0x06, 0x09, 0x60, 0x86, 0x48, 0x01, 0x65, 0x03, 0x04, 0x02,
-                0x03, 0x05, 0x00, 0x04, 0x40,
-            ],
-
-            // A special TLS case which doesn't use an ASN1 prefix
-            Hash::MD5SHA1 => &[],
```
newpavlov, line 65: What about this case? Or we simply say that it's insecure and not supported by this crate directly?

tarcieri: I'd be fine with that personally

Cargo.toml
lumag: I've added the sha-3 test

lumag:
@tarcieri @newpavlov Note, this pull request contains breaking change, it drops rsa::Hash. If you wish I can pull this commit out  and leave only non-breaking changs.

src/hash.rs
lumag: I'd say, it should be implemented as a separate `Digest` wrapping MD5+SHA1 internally.

tarcieri:
@lumag breaking changes are fine as there are already ones merged (`master` is currently v0.7.0-pre)

tarcieri (approved):

newpavlov:
It would be nice to update the changelog. It's easier to edit "unreleased" section, than to collect all changes introduced since previous release.

tarcieri:
@newpavlov unfortunately there is no changelog! (#151)

But I'm happy to add one and document all of the changes in this (upcoming) release.

tarcieri:
I'm going to go ahead and merge this as there are several people asking for an RC.

Can wait for feedback from @dignifiedquire before cutting a final release. And in the meantime, I can get a changelog added.

newpavlov:
>unfortunately there is no changelog

Ah, true. I think the next breaking release is  a good opportunity to add one then.

dignifiedquire:
Uhm, I am not sure I can support this, but from first glance this means I can't change the hash function at runtime, which is functionality I need. Or am I missing something?

src/pkcs1v15.rs
```diff
@@ -741,22 +783,22 @@ mod tests {
...
-                    "02d42875a7ffa4a148aa311f9e4b562e4e13a2223fb15f4e5bf5f2b206d9451c"
+                    "a4f3fa6ea93bcdd0c57be020c1193ecbfd6f200a3d95c409769b029578fa0e33"
+                    "6ad9a347600e40d3ae823b8c7e6bad88cc07c1d54c3a1523cbbb6d58efc362af"
                 ),
                 false,
             ),
         ];
         let pub_key: RsaPublicKey = priv_key.into();
-        let verifying_key = VerifyingKey::<Sha256>::new_with_prefix(pub_key);
+        let verifying_key = VerifyingKey::<Sha1>::new_with_prefix(pub_key);
```
dignifiedquire, line 801: why did these get changed?

lumag: It was a revert of previous commit which changed Sha1 to Sha256.

lumag:
> Uhm, I am not sure I can support this, but from first glance this means I can't change the hash function at runtime, which is functionality I need. Or am I missing something?

Unfortunately changing the hash function at runtime doesn't match the `SigningKey` model. You can have several signing keys at once.

tarcieri:
@dignifiedquire we could ensure the low-level APIs (defined in terms of `RsaPrivateKey`/`RsaPublicKey` operate in terms of `DynDigest` and accept an `ObjectIdentifier` parameter where necessary. Does that work?

Re: `SigningKey`, using (Randomized)`DigestSigner` would at least reduce the part that needs to be monomorphized to the signing function itself, which would call into a common implementation once the digest has been computed.

An enum could be used to select the concrete `Digest` to use when invoking (Randomized)`DigestSigner`, although that could potentially be defined to represent a specific ciphersuite in a downstream crate, rather than in the `rsa` crate itself.

lumag:
@dignifiedquire The `SigningKey` expects to get clear text, not the pre-hashed data. `DigestSigningKey` accepts `Digest` instance. I can implement the hazmat's interfaces to be able to accept prehashed texts. Or the old API is still accessible.

dignifiedquire:
This is the code that I need to support: https://github.com/rpgp/rpgp/blob/master/src/crypto/rsa.rs#L67-L82 
The hash function is supplied dynamically with a list of supported hashes here: https://github.com/rpgp/rpgp/blob/master/src/crypto/hash.rs#L45

dignifiedquire:
> I can implement the hazmat's interfaces to be able to accept prehashed texts

This specifically needs to support prehashed input. The linked api usage is 80% of the reason that I originally wrote this library. 😅

lumag:
@dignifiedquire ack, I see the issue. I'll send a pull request in one of the forthcoming days.

dignifiedquire:
thanks @lumag!

lumag:
@dignifiedquire I've sketched https://github.com/rpgp/rpgp/pull/193. However it might be easier to just restore the rsa::Hash struct and use old API as you have been using it up to now.

PR #185, opened 2022-09-10
Drop reference boilerplate code

lumag:
Drop boilerplate code which implements traits for references to key types. The code can use references directly.


tarcieri (approved):
This seems fine to me but I'm curious what @dignifiedquire thinks.

also cc @str4d

PR #190, opened 2022-09-16
Implement Keypair trait for the RSA keys

lumag:
This is rather intrusive implementation. It reworks the way `RsaPrivateKey` is organized. Unfortunately it also meant that default `Serialize` an `Deserilialize` derivations provide a list of tokens that are not backwards-compatible. Thus I had to implement those traits manually.

### DIFF
--- a/src/encoding.rs
+++ b/src/encoding.rs
@@ -64,56 +64,70 @@ impl DecodePublicKey for RsaPublicKey {}
 
 impl EncodePrivateKey for RsaPrivateKey {
     fn to_pkcs8_der(&self) -> pkcs8::Result<SecretDocument> {
-        // Check if the key is multi prime
-        if self.primes().len() > 2 {
-            return Err(pkcs1::Error::Version.into());
-        }
-
-        let modulus = self.n().to_bytes_be();
-        let public_exponent = self.e().to_bytes_be();
-        let private_exponent = Zeroizing::new(self.d().to_bytes_be());
-        let prime1 = Zeroizing::new(self.primes()[0].to_bytes_be());
-        let prime2 = Zeroizing::new(self.primes()[1].to_bytes_be());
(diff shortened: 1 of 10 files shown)


tarcieri:
@lumag I'm confused why you need to change `RsaPublicKey`/`RsaPrivateKey` at all

lumag:
> @lumag I'm confused why you need to change `RsaPublicKey`/`RsaPrivateKey` at all

@tarcieri because I have to embed a `VerifyingKey` as a field into the `SigningKey`. Otherwise `as_ref()` will fail with the _returning a reference for temp variable_ error. And putting `VerifyingKey` next to `RsaPrivateKey` results in information duplication. So...I ended up with `RsaPrivateKeyParts`,

tarcieri:
@lumag I opened a tracking issue for problems with `Keypair` and suggested a possible alternative: https://github.com/RustCrypto/traits/issues/1124

lumag:
@tarcieri as a side note:
While this PR is intrusive, it removes the required usage of `RsaPrivateKey`, making all `SigningKey` implementations first class citizens. Also note, that ECDSA signing keys also incorporate `VerifyingKey` as a field.

lumag:
I have rebased this PR on top of RSA 0.7.0. Note, I had to disable `AsRef<RsaPrivateKey> for SigningKey<D>` implementations. Since `RsaPrivateKey` is not a field of `SigningKey`, I don't see a way to implement `AsRef`

tarcieri:
@lumag PTAL at the `Keypair` changes here:

https://github.com/RustCrypto/traits/pull/1141/files#diff-1eb80c5452df3ad684ce36b4bfb4590a600978c10f0135bcba713ca18617d9bc

It makes the `AsRef` support an optional optimization, and changes the default `Keypair` trait to return the public key by value.

PR #206, opened 2022-10-07
Add `RsaPublicKey::new_unchecked`

### LINKED ISSUE #199: Having a `MAX_PUB_EXPONENT` makes this crate unusable with some applications
Some applications use very large numbers, including ones from [factoring challenges](https://en.wikipedia.org/wiki/RSA_numbers), for the exponent part of the key.

Having a `MAX_PUB_EXPONENT` makes this crate unusable with such applications without using a patched source tree.

If the concern is compatibility with platforms that do enforce such limits, then the limit should be optional, and the purpose of it should be stated explicitly, IMHO.

tarcieri:
Constructor for `RsaPublicKey` which bypasses all checks around the modulus and public exponent size.

Closes #199.

### DIFF
--- a/src/key.rs
+++ b/src/key.rs
@@ -250,9 +250,19 @@ impl RsaPublicKey {
 
     /// Create a new public key from its components.
     pub fn new_with_max_size(n: BigUint, e: BigUint, max_size: usize) -> Result<Self> {
-        let k = RsaPublicKey { n, e };
+        let k = Self { n, e };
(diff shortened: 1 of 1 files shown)


tarcieri:
Hopefully this is uncontroversial enough I'm just going to merge it.

However, I'll hold off on approval before a final release.

dignifiedquire:
Thanks, looks good to me 👍

PR #215, opened 2022-10-20
padding: ensure PaddingScheme is Send and Sync

### LINKED ISSUE #214: PaddingScheme being !Send makes things harder
See https://github.com/librespot-org/librespot/issues/1065

`PaddingScheme` is `!Send` making it impossible to use it in async code requiring `Send` futures.

Is there any way this could be prevented?

gdesmott:
It's more convenient for users.

Fix #214

### DIFF
--- a/src/key.rs
+++ b/src/key.rs
@@ -930,7 +930,9 @@ mod tests {
         }
     }
 
-    fn do_test_encrypt_decrypt_oaep<D: 'static + Digest + DynDigest>(prk: &RsaPrivateKey) {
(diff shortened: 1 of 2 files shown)


gdesmott:
Do we want `Sync` as well?

gdesmott:
> Do we want `Sync` as well?

Added it as it's free and more convenient.

tarcieri:
This is technically a breaking change, although probably one which won't lead to breakages in practice

tarcieri (approved):

PR #216, opened 2022-10-29
Add missing rustdoc comments; enable `missing_docs` lint

tarcieri:
Several types and methods were missing documentation.

This commit adds document and enables warnings for `missing_docs`.

Additionally it updates all references to PKCS#1 RFCs to use RFC8017, which documents the latest version of PKCS#1.

cc @lumag

### DIFF
--- a/src/errors.rs
+++ b/src/errors.rs
@@ -1,26 +1,64 @@
+//! Error types.
+
+/// Alias for [`core::result::Result`] with the `rsa` crate's [`Error`] type.
 pub type Result<T> = core::result::Result<T, Error>;
 
 /// Error types
 #[derive(Debug, Eq, PartialEq)]
 #[non_exhaustive]
 pub enum Error {
+    /// Invalid padding scheme.
     InvalidPaddingScheme,
+
+    /// Decryption error.
     Decryption,
+
+    /// Verification error.
     Verification,
+
+    /// Message too long.
     MessageTooLong,
+
+    /// Input must be hashed.
     InputNotHashed,
+
+    /// Number of primes must be 2 or greater.
     NprimesTooSmall,
+
+    /// Too few primes of a given length to generate an RSA key.
     TooFewPrimes,
+
+    /// Invalid prime value.
     InvalidPrime,
+
+    /// Invalid modulus.
     InvalidModulus,
+
+    /// Invalid exponent.
     InvalidExponent,
+
+    /// Invalid coefficient.
     InvalidCoefficient,
+
+    /// Modulus too large.
     ModulusTooLarge,
+
+    /// Public exponent too small.
     PublicExponentTooSmall,
+
+    /// Public exponent too large.
     PublicExponentTooLarge,
+
+    /// PKCS#1 error.
     Pkcs1(pkcs1::Error),
+
+    /// PKCS#8 error.
     Pkcs8(pkcs8::Error),
+
+    /// Internal error.
     Internal,
+
+    /// Label too long.
     LabelTooLong,
 }
 
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -1,9 +1,17 @@
 //! RSA Implementation in pure Rust.
 //!
+//! It supports several schemes described in [RFC8017]:
+//!
+//! - OAEP encryption scheme
+//! - PKCS#1 v1.5 encryption scheme
+//! - PKCS#1 v1.5 signature scheme
+//! - PSS signature scheme
+//!
+//! These schemes are described below.
 //!
 //! # Usage
 //!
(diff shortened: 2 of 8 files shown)


src/errors.rs
```diff
@@ -1,26 +1,62 @@
+/// Alias for [`core::result::Result`] with the `rsa` crate's [`Error`] type.
 pub type Result<T> = core::result::Result<T, Error>;
 
 /// Error types
 #[derive(Debug, Eq, PartialEq)]
 #[non_exhaustive]
 pub enum Error {
+    /// Invalid padding scheme.
```
tarcieri, line 8: Alternatively we could disable the lint for this type, if you think the variant names are self-explanatory.

src/lib.rs
```diff
@@ -183,10 +183,14 @@
 //! # Ok(())
 //! # }
 //! ```
+//!
+//! [`pkcs8::DecodePublicKey`]: https://docs.rs/pkcs8/latest/pkcs8/trait.DecodePublicKey.html
+//! [`pkcs8::EncodePublicKey`]: https://docs.rs/pkcs8/latest/pkcs8/trait.EncodePublicKey.html
```
tarcieri, line 188: For some reason rustdoc is failing to link these. Not sure why, but this at least works around the bug.

See: https://docs.rs/rsa/latest/rsa/#pkcs8-rsa-key-encoding

src/lib.rs
```diff
@@ -202,9 +206,17 @@ pub mod algorithms;
...
 pub mod pkcs1v15;
-/// RSASSA-PSS Signature support
+/// Support for the [Probabilistic Signature Scheme] (PSS) designed by
+/// Mihir Bellare and Phillip Rogaway, a.k.a. RSASSA-PSS.
+///
+/// Specified in [RFC3447 § 8.1].
+///
+/// [Probabilistic Signature Scheme]: https://en.wikipedia.org/wiki/Probabilistic_signature_scheme
+/// [RFC8017 § 8.1]: https://datatracker.ietf.org/doc/html/rfc8017#section-8.1
 pub mod pss;
```
tarcieri, line 220: Might be worth moving all of these comments to be toplevel rustdoc comments on the respective modules, rather than defined inline here.

src/errors.rs
lumag: Nah, let's not disable the linter.

src/lib.rs
lumag: Strange. This works for me with the local cargo doc.

src/lib.rs
```diff
@@ -202,9 +206,17 @@ pub mod algorithms;
...
-/// RSASSA-PKCS1-v1_5 Signature support
+/// PKCS#1 v1.5 support as described in [RFC8017 § 8.2].
+///
+/// [RFC8017 § 8.2]: https://datatracker.ietf.org/doc/html/rfc8017#section-8.2
 pub mod pkcs1v15;
-/// RSASSA-PSS Signature support
+/// Support for the [Probabilistic Signature Scheme] (PSS) designed by
+/// Mihir Bellare and Phillip Rogaway, a.k.a. RSASSA-PSS.
+///
+/// Specified in [RFC3447 § 8.1].
```
lumag, line 216: RFC8017 #  8.1

src/lib.rs
tarcieri: Oops, just noticed this comment cites the old RFC too

src/lib.rs
tarcieri: Yup, good catch

src/lib.rs
lumag: No strong opinion. Maybe it's worth adding mentions of PKCS#1 v1.5, PSS and OAEP to the top-level description (so that the developer can see what's supported), while keeping such details and links here, in the inline comments.

tarcieri: rustdoc automatically puts the first line in the toplevel description, regardless of where it’s defined

lumag: I meant something like:
```diff
-//! RSA Implementation in pure Rust.
+//! RSA Implementation in pure Rust. It supports several schemes described in RFC 8017:
+//! - PKCS#1 v1.5 encryption scheme
+//! - OAEP encryption scheme
+//! - PKCS#1 v1.5 signature scheme
+//! - PSS signature scheme
 //!
+//! These schemes are described below.
 //!
 //! # Usage
 //!
```

tarcieri: Sure, that works.

tarcieri: Added that and moved the other comments to the tops of the respective modules.

lumag:
LGTM

PR #217, opened 2022-10-30
feat: switch to version 2.0 (pre) of the  signature crate

lumag:
Rework the crate to implement traits from the preview of the signature crate. Use Vec<u8> as Self::Repr type.

Note: I'm yet to see how will this impact the code in my projects. But for the evaluation I'd first need `ecdsa` / `p384` to be updated.

cc @tarcieri @dignifiedquire

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -17,13 +17,13 @@ num-bigint = { version = "0.8.1", features = ["i128", "u64_digit", "prime", "zer
 num-traits = { version= "0.2.9", default-features = false, features = ["libm"] }
 num-integer = { version = "0.1.39", default-features = false }
 num-iter = { version = "0.1.37", default-features = false }
-rand_core = { version = "0.6", default-features = false }
+rand_core = { version = "0.6.4", default-features = false }
 byteorder = { version = "1.3.1", default-features = false }
 subtle = { version = "2.1.1", default-features = false }
 digest = { version = "0.10.5", default-features = false, features = ["alloc", "oid"] }
 pkcs1 = { version = "0.4", default-features = false, features = ["pkcs8", "alloc"] }
 pkcs8 = { version = "0.9", default-features = false, features = ["alloc"] }
-signature = { version = "1.6.4", default-features = false , features = ["digest-preview", "rand-preview"] }
+signature = { version = "2.0.0-pre.2", default-features = false , features = ["digest-preview", "rand-preview"] }
 zeroize = { version = "1", features = ["alloc"] }
 
 # Temporary workaround until https://github.com/dignifiedquire/num-bigint/pull/42 lands
@@ -53,7 +53,7 @@ name = "key"
 
 [features]
 default = ["std", "pem"]
-hazmat = ["signature/hazmat-preview"]
+hazmat = []
 nightly = ["num-bigint/nightly"]
 serde = ["num-bigint/serde", "serde_crate"]
 expose-internals = []
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "rsa"
-version = "0.7.2"
+version = "0.8.0-pre"
 authors = ["RustCrypto Developers", "dignifiedquire <dignifiedquire@gmail.com>"]
 edition = "2021"
 description = "Pure Rust RSA implementation"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -53,7 +53,6 @@ name = "key"
 
 [features]
 default = ["std", "pem"]
-hazmat = []
 nightly = ["num-bigint/nightly"]
 serde = ["num-bigint/serde", "serde_crate"]
 expose-internals = []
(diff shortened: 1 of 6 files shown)


lumag:
Converted to draft to prevent merging the PR with `[patch.crates-io]` in `Cargo.toml`

tarcieri:
FYI, I'm cutting a crate prerelease here shortly: https://github.com/RustCrypto/traits/pull/1145

lumag:
@dignifiedquire @tarcieri gracious ping for the review.

Cargo.toml
```diff
@@ -53,7 +53,7 @@ name = "key"
 
 [features]
 default = ["std", "pem"]
-hazmat = ["signature/hazmat-preview"]
+hazmat = []
```
tarcieri, line 56: Can probably remove the feature gating here entirely

src/pkcs1v15.rs
```diff
@@ -20,60 +19,52 @@ use crate::errors::{Error, Result};
 use crate::key::{self, PrivateKey, PublicKey};
 use crate::{RsaPrivateKey, RsaPublicKey};
 
-#[derive(Clone)]
+#[derive(Clone, PartialEq, Eq)]
 pub struct Signature {
-    bytes: Vec<u8>,
+    bytes: Box<[u8]>,
```
tarcieri, line 24: Though this is ok as a first pass, I think ideally you could replace the internal representation with a `BigUint` now, which would eliminate the need for an intermediate `Box<[u8]>` when signing/verifying.

tarcieri:
@lumag can you rebase?

Cargo.toml
lumag: Well, for me the goal was to prevent hazmat interfaces from being available by default. I can drop it of course.

src/pkcs1v15.rs
lumag: Ack, I'll take a look on it later.

lumag:
@tarcieri rebased

tarcieri (approved):
This looks fine now as a first pass.

Curious what @dignifiedquire thinks about another breaking release so soon.

We should coordinate with `ed25519-dalek` and the https://github.com/RustCrypto/signatures crates to have `signature` v2.0 releases ready at the same time.

Cargo.toml
```diff
@@ -23,7 +23,7 @@ subtle = { version = "2.1.1", default-features = false }
 digest = { version = "0.10.5", default-features = false, features = ["alloc", "oid"] }
 pkcs1 = { version = "0.4", default-features = false, features = ["pkcs8", "alloc"] }
 pkcs8 = { version = "0.9", default-features = false, features = ["alloc"] }
-signature = { version = "1.6.4", default-features = false , features = ["digest-preview", "rand-preview"] }
+signature = { version = "2.0.0-pre", default-features = false , features = ["digest-preview", "rand-preview"] }
```
tarcieri, line 26: Can you switch to the v2.0.0-pre.2 crate release? https://crates.io/crates/signature/2.0.0-pre.2

lumag: Done

src/pkcs1v15.rs
lumag: Opened #220 . Implementing it would require changing the internals of the whole RSA crate.

tarcieri: Thanks

dignifiedquire:
> Curious what @dignifiedquire thinks about another breaking release so soon.

- I think we should put this out as a prerelease, and only fully release it once we are sure the other crates using the signature crates are going to be updated/are already updated to avoid fragmentation and pain in upstream use cases
- we should gather the biggest users of the library and make sure they can easily upgrade (I only know about `rpgp`, but expect/hope there are others out there)

tarcieri:
@dignifiedquire yeah, the plan is to coordinate with `ed25519-dalek` so we can release all of the major crates which use `signature` at the same time.

I think everything on `master` has been released so we could potentially merge this (after a rebase) and cut a prerelease after that

tarcieri:
@lumag can you rebase plase?

lumag:
> @lumag can you rebase please?

Done

src/pkcs1v15.rs
```diff
@@ -342,12 +333,13 @@ where
...
     }
 
-    pub(crate) fn key(&self) -> &RsaPrivateKey {
-        &self.inner
-    }
-
-    pub(crate) fn prefix(&self) -> Vec<u8> {
-        self.prefix.clone()
+    /// Generate a new signing key.
+    pub fn random<R: CryptoRngCore>(rng: &mut R, bit_size: usize) -> Result<Self> {
```
tarcieri, line 337: Based on discussion in https://github.com/RustCrypto/traits/issues/1148 we're leaning towards:

```suggestion
    pub fn random<R: CryptoRngCore + ?Sized>(rng: &mut R, bit_size: usize) -> Result<Self> {
```

tarcieri:
@lumag looks good other than the one nit about the RNG trait bounds

lumag:
@tarcieri Added the `?Sized` where it seems suitable.

tarcieri:
The PRs for updating the `ecdsa`, `ed25519`, `ed25519-dalek`, and `p256`/`p384`/`k256` crates have all landed.

This is the only remaining one. I'd suggest we merge this soon so we can further iterate on `signature` 2.0 work.

PR #221, opened 2022-11-08
Make PrecomputedValues public

BrettMayson:
I need access to these values in order to store a custom private key format, this is preventing me from switching away from openssl

### DIFF
--- a/src/key.rs
+++ b/src/key.rs
@@ -103,19 +103,19 @@ impl Deref for RsaPrivateKey {
 }
 
 #[derive(Debug, Clone)]
-pub(crate) struct PrecomputedValues {
+pub struct PrecomputedValues {
     /// D mod (P-1)
-    pub(crate) dp: BigUint,
+    pub dp: BigUint,
     /// D mod (Q-1)
-    pub(crate) dq: BigUint,
+    pub dq: BigUint,
     /// Q^-1 mod P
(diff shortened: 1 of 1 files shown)


tarcieri:
Is there a particular reason why you want to store these values instead of computing them on-the-fly?

The trend in RSA private key format design has been to store fewer precomputed values rather than more. Modern key formats store only `d` and can compute everything else from that and the public key.

BrettMayson:
I need to store a custom format that I have no control over, which requires these values to be stored

tarcieri:
Do you actually need to support multiprime RSA?

It would probably be worth keeping that out of the public API, especially as it currently has an odd definition owing to PKCS#1 also needing the CRT coefficient for two primes, so the current vector stores "additional CRT coefficients" beyond that one IIUC.

newpavlov:
I don't think we should make the fields public. If you only need to read pre-computed values, then a bunch of getters should do the job. If you need to restore the values from serialized data, then we would need to think about validation.

BrettMayson:
> Also: are you actually using multiprime RSA? (re: crt_values)

No, just `dp`, `dq` and `qinv`

tarcieri:
In that case it sounds like you should just add read-only getter methods for those three

tarcieri (approved):

dignifiedquire (approved):

PR #233, opened 2022-11-30
private_key encrypt keep pace with gorsa

Caisin:
this crate use publickey encrypt msg result is not eq other language(go:gorsa,etc.)


tarcieri:
This looks like you're trying to implement PKCS#1 v1.5 padding? That's already definitely supported.

https://github.com/RustCrypto/RSA/blob/35f7ae2/src/pkcs1v15.rs#L108-L132

Here's the public API:

https://docs.rs/rsa/0.7.2/rsa/padding/enum.PaddingScheme.html#method.new_pkcs1v15_encrypt

Caisin:
https://github.com/RustCrypto/RSA/blob/35f7ae2/src/pkcs1v15.rs#L108-L132

tarcieri:
> can't encrypt long str

It sounds like you'd like to use RSA in ECB mode to encrypt data larger than the modulus. That's something of an antipattern. A better solution is to encrypt a symmetric key, and then use the symmetric key to encrypt the data, a.k.a. [hybrid cryptography](https://en.wikipedia.org/wiki/Hybrid_cryptosystem).

You can open an issue for ECB mode support if you'd like.

> use this function encrypt result,other language can't decrypt.

Please provide complete code examples in both languages. If I had to guess it's probably a bug somewhere in your code, rather than this library, which is widely used and tested.

Caisin:
i create a issue at #236

Issue #226, opened 2022-11-15
Should `PaddingScheme` be a trait instead of an enum?


tarcieri:
Looking at issues like #215 it seems like requirements of certain variants are making the others somewhat harder to use.

The `PKCS1v15Encrypt` and `PKCS1v15Sign` have relatively minimal requirements with `PKCS1v15Encrypt` having no members whatsoever, whereas `OAEP` and `PSS` both need `Box<dyn DynDigest + Send + Sync>`.

Having a trait would allow the impls to live in e.g. `pkcs1v15` and `pss` modules as well, whereas right now the `pkcs1v15` module only exposes encryption-related functionality.

newpavlov:
Maybe it's better to completely remove `PaddingScheme` and instead introduce separate padding-specific methods on `RsaPrivateKey`?

tarcieri:
@newpavlov both the public and private key need padding-related functionality. The public key needs to generate a padded message for encryption, and verify the padding with a signature, with the inverse true for private keys.

lumag:
I'd second @newpavlov here. I'd drop the `PaddingScheme` completely instead. For the sign/verify ops we have `Signer` / `Verifer` families which do not need `PaddingScheme`. So, I think, it would be logical to add `Encryptor` / `Decryptor` and drop the PS.

lumag:
@tarcieri I'd start with something like https://github.com/RustCrypto/RSA/pull/245

tarcieri:
It's going to be a nightmare to rebase #244 on that :cry:

It would've been better if you started with #244 as the base.

lumag:
@tarcieri well, I branched off the trunk on purpose. IMHO there is no point in having `SignatureScheme` if we can drop it first.

lumag:
@tarcieri @newpavlov @dignifiedquire  Also I think that we should make encryption/decryption follow the `Signer` API. In other words, have a separate `Encryptor` / `Decryptor` traits, implement padding-specific keys and drop `PaddingScheme` completely.

lumag:
I implemented my ideas as #246

tarcieri:
Resolved by #244

PR #244, opened 2023-01-06
Refactor `PaddingScheme` into a trait

tarcieri:
As proposed in #226, splits up the `PaddingScheme` enum into four structs, named after the previous variants of the struct (adopting capitalization from the Rust API guidelines):

- `oaep::Oaep`
- `pkcs1v15::{Pkcs1v15Encrypt, Pkcs1v15Sign}`
- `pss::Pss`

All of these are re-exported from the toplevel.

Each of these structs impls one or more of the following traits:

- `PaddingScheme`: used for encryption
- `SignatureScheme`: used for signing

The `PaddingScheme` constructors have been remapped as follows:

- `new_oaep` => `Oaep::new`
- `new_oaep_with_label` => `Oaep::new_with_label`
- `new_oaep_with_mgf_hash` => `Oaep::new_with_mgf_hash`
- `new_oaep_with_mgf_hash_with_label` => `Oaep::new_with_mgf_hash_and_label`
- `new_pkcs1v15_encrypt` => `Pkcs1v15Encrypt`
- `new_pkcs1v15_sign` => `Pkcs1v15Sign::new`
- `new_pkcs1v15_sign_raw` => `Pkcs1v15Sign::new_raw`
- `new_pss` => `Pss::{new, new_blinded}`
- `new_pss_with_salt` => `Pss::{new_with_salt new_blinded_with_salt}`

### DIFF
--- a/src/key.rs
+++ b/src/key.rs
@@ -172,10 +172,10 @@ impl From<&RsaPrivateKey> for RsaPublicKey {
 /// Generic trait for operations on a public key.
 pub trait PublicKey: EncryptionPrimitive + PublicKeyParts {
     /// Encrypt the given message.
-    fn encrypt(
+    fn encrypt<R: CryptoRngCore, P: PaddingScheme>(
         &self,
-        rng: &mut impl CryptoRngCore,
-        padding: impl PaddingScheme,
+        rng: &mut R,
+        padding: P,
         msg: &[u8],
     ) -> Result<Vec<u8>>;
 
@@ -185,7 +185,7 @@ pub trait PublicKey: EncryptionPrimitive + PublicKeyParts {
     /// passed in through `hash`.
     ///
     /// If the message is valid `Ok(())` is returned, otherwise an `Err` indicating failure.
-    fn verify(&self, padding: impl SignatureScheme, hashed: &[u8], sig: &[u8]) -> Result<()>;
+    fn verify<S: SignatureScheme>(&self, scheme: S, hashed: &[u8], sig: &[u8]) -> Result<()>;
 }
 
 impl PublicKeyParts for RsaPublicKey {
@@ -199,17 +199,17 @@ impl PublicKeyParts for RsaPublicKey {
 }
 
 impl PublicKey for RsaPublicKey {
-    fn encrypt(
+    fn encrypt<R: CryptoRngCore, P: PaddingScheme>(
         &self,
-        rng: &mut impl CryptoRngCore,
-        padding: impl PaddingScheme,
+        rng: &mut R,
+        padding: P,
         msg: &[u8],
     ) -> Result<Vec<u8>> {
         padding.encrypt(rng, self, msg)
     }
 
-    fn verify(&self, padding: impl SignatureScheme, hashed: &[u8], sig: &[u8]) -> Result<()> {
-        padding.verify(self, hashed, sig)
+    fn verify<S: SignatureScheme>(&self, scheme: S, hashed: &[u8], sig: &[u8]) -> Result<()> {
+        scheme.verify(self, hashed, sig)
     }
 }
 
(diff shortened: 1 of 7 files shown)


tarcieri:
cc @lumag

tarcieri:
Note that there's various additional refactoring that could probably be performed, like getting rid of the inherent methods for these functionality and using the traits exclusively, or getting rid of the `SignatureScheme` trait and factoring the relevant code into `SigningKey` and `VerifyingKey`.

I deliberately didn't make any of those kinds of changes to make code review easier.

src/key.rs
```diff
@@ -198,36 +199,17 @@ impl PublicKeyParts for RsaPublicKey {
 }
 
 impl PublicKey for RsaPublicKey {
-    fn encrypt<R: CryptoRngCore>(
+    fn encrypt(
         &self,
-        rng: &mut R,
-        padding: PaddingScheme,
+        rng: &mut impl CryptoRngCore,
+        padding: impl PaddingScheme,
```
dignifiedquire, line 205: Why did you go for `impl` over a generic? I have been bitten by libraries which used `impl` because you can't provide type hints directly, so I am sceptical on this

src/oaep.rs
```diff
@@ -1,20 +1,169 @@
...
+            self.label,
+        )
+    }
+}
+
+impl fmt::Debug for Oaep {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        f.debug_struct("OAEP")
+            .field("label", &self.label)
+            .finish_non_exhaustive()
```
dignifiedquire, line 163: this is a bit odd, not listing all fields and having the struct not actually be non exhaustive, while printing it as such

dignifiedquire:
I overall very much like this approach, it makes the API clearer to understand and reason about.

src/key.rs
tarcieri: I can add the explicit generic parameters back, however it is possible to type hint with the arguments rather than the generic parameters:

https://github.com/RustCrypto/RSA/pull/244/files#diff-0988d5d2259e1a401001675578c38d09c46290b877b66b07fb96a7e034ccdf70R434

src/oaep.rs
tarcieri: Well, the old impl looked like this...

```rust
            PaddingScheme::OAEP { ref label, .. } => {
                // TODO: How to print the digest name?
                write!(f, "PaddingScheme::OAEP({:?})", label)
            }
```

It's at least a step up from that. I suppose it could list out the other field names, but we have nothing to display for them.

dignifiedquire: Yeah, previous one wasn't great either. What I usually do is add the fields and print the type description or sth like that, so that at least there is some info on what is there.

src/key.rs
dignifiedquire: but that only works if working with `Option` right?

tarcieri: You can always use `as` as a sort of pseudo type ascription as well, e.g.:

```rust
None as Option<&mut DummyRng>
```

works in addition to

```rust
Option::<&mut DummyRng>::None
```

...but I can go ahead and change the `impl Trait` usages back to explicit parameters

tarcieri: befaa734e06c78a9f8c686616effc0cd36b902af switches things back to generic parameters

src/oaep.rs
tarcieri: [69c9ba0](https://github.com/RustCrypto/RSA/pull/244/commits/69c9ba0e91a0684fb0f2d38d364e09d617097cee) adds the field names.

Perhaps we could add a `DynDigestWithOid` trait to `Digest` which would make it possible to get the OID for the digest, but for now I just used `...`

tarcieri:
@dignifiedquire this should be good to go now

src/key.rs
```diff
@@ -173,18 +172,20 @@ impl From<&RsaPrivateKey> for RsaPublicKey {
 /// Generic trait for operations on a public key.
 pub trait PublicKey: EncryptionPrimitive + PublicKeyParts {
     /// Encrypt the given message.
-    fn encrypt<R: CryptoRngCore>(
+    fn encrypt(
         &self,
-        rng: &mut R,
-        padding: PaddingScheme,
+        rng: &mut impl CryptoRngCore,
+        padding: impl PaddingScheme,
```
dignifiedquire, line 178: don't we want to normalize on `impl` vs generics?

src/key.rs
dignifiedquire: @tarcieri I am confused, the latest shows `impl` again

src/key.rs
```diff
@@ -448,113 +430,37 @@ impl RsaPrivateKey {
...
-            } => pss::sign::<R, _>(rng, false, self, digest_in, salt_len, &mut *digest),
-            _ => Err(Error::InvalidPaddingScheme),
-        }
-    }
-
-    /// Sign the given digest.
-    ///
-    /// Use `rng` for blinding.
-    pub fn sign_blinded<R: CryptoRngCore>(
+    pub fn sign_with_rng(
```
dignifiedquire, line 457: Can you add some clearer docs on when signatures are now blinded?

src/key.rs
tarcieri: Oh sorry, https://github.com/RustCrypto/RSA/commit/befaa734e06c78a9f8c686616effc0cd36b902af fixed some of the cases but not all of them. Will do another pass.

src/key.rs
tarcieri: [dc47326](https://github.com/RustCrypto/RSA/pull/244/commits/dc473264d16b6c59969b3fb1d287c50f4483ac34) should change the remaining usages of `impl Trait` in the PR (back) to generic parameters

src/key.rs
tarcieri: "Blinding" is somewhat conflated between random blinding to mask private key operations and RSA-BSSA Blind Signatures.

Everything related to the latter is now under the `pss` module, namely `Pss:new_blinded` and `Pss::new_blinded_with_salt`, along with `BlindedSigningKey`.

I guess I can spell all of that out in the comment?

tarcieri: [f39bf60](https://github.com/RustCrypto/RSA/pull/244/commits/f39bf608b1a0693a285b93396ddaa2087b8154b3) adds more documentation to `sign_with_rng` including information on how to use it with RSA-BSSA blind signatures

dignifiedquire (approved):
thank you, lgtm

PR #245, opened 2023-01-06
feat: drop old signing/verification API

lumag:
The signing and verification is now implemented using the generic Signature and *Signer / *Verifier traits. Drop old API proprietary to the RSA crate.

### DIFF
--- a/benches/key.rs
+++ b/benches/key.rs
@@ -6,6 +6,7 @@ use base64ct::{Base64, Encoding};
 use num_bigint::BigUint;
 use num_traits::{FromPrimitive, Num};
 use rand_chacha::{rand_core::SeedableRng, ChaCha8Rng};
+use rsa::signature::RandomizedSigner;
 use rsa::{PaddingScheme, RsaPrivateKey};
 use sha2::{Digest, Sha256};
 use test::Bencher;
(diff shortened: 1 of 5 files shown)


benches/key.rs
```diff
@@ -41,17 +42,25 @@ fn bench_rsa_2048_pkcsv1_decrypt(b: &mut Bencher) {
 #[bench]
 fn bench_rsa_2048_pkcsv1_sign_blinded(b: &mut Bencher) {
     let priv_key = get_key();
+    let signing_key = rsa::pkcs1v15::SigningKey::<Sha256>::new_with_prefix(priv_key);
```
dignifiedquire, line 45: I think this is a regression in API usability: having to create a newtype struct to call a single function.

lumag: @dignifiedquire why? The signing_key is a long term object, which can be stored and further used. This is more or less what we have for other PK types. Also I think we have all API in place to work with these keys as a first-class objects (generate, read, write, etc). The only thing that prevents me from thinking about turning `RsaPublicKey` and `RsaPrivateKey` into create-private structs is the existence of S+E keys.

dignifiedquire: Because now I am forced to decide on the padding + hashing when storing the key. Most of my use cases are that these are dynamically chosen parameters, so I need to store the key itself, and apply the different algorithms depending on runtime inputs.

dignifiedquire: >  This is more or less what we have for other PK types

Yes, and I think it is not great there either, if you have dynamic cases.

dignifiedquire: as an example, the code I would have to write now will look sth like this

```rust
match padding {
  "pkcs" => {
    let key = rsa::pkcs1v15::SigningKey::<Sha256>::new_with_prefix(priv_key);
    key.sign()
  }
  ...
}
```

tarcieri: @lumag I think it's somewhat more common with RSA due to legacy reasons for keys to be used in multiple roles, even across encryption and signing.

See for example TLS < 1.3 using RSA keys for both key encipherment and digital signatures as part of a handshake.

Which algorithm is used depends on a combination of server-side configuration and the capabilities of the client.

lumag: Yeah. Or the (rare) GPG S+E keys.
So, how do we want to proceed? Do we drop these two PRs (perfectly fine with me) or would you like for me to improve them? Let's probably decide on this one first.

tarcieri:
Seems like we're keeping this API for now

PR #246, opened 2023-01-06
Remove padding using new encryption API

lumag:
This is a reimplementation of RSA encryption/decryption API on top of #245 .

cc @tarcieri @newpavlov @dignifiedquire


lumag:
This PR incorporates #245  instead of replacing it to let us discuss these two PRs separately.

src/lib.rs
```diff
@@ -13,7 +13,9 @@
 //!
 //! ## PKCS#1 v1.5 encryption
 //! ```
-//! use rsa::{PublicKey, RsaPrivateKey, RsaPublicKey, PaddingScheme};
+//! use rsa::{RsaPrivateKey, RsaPublicKey};
+//! use rsa::pkcs1v15::{DecryptingKey, EncryptingKey};
```
dignifiedquire, line 17: `EncryptingKey` reads quite odd, I would have expected it to be called `EncryptionKey`

lumag: Ack

tarcieri:
I do like this approach overall.

I worry there's going to be a lot of complaints and confusion for anyone attempting to upgrade who just wants to be able to call an inherent method on a key type.

If we do go with this I think it needs some documentation in the changelog or elsewhere about how to upgrade from the previous API.

tarcieri:
Re: this vs #244, I think both would be interesting?

I think there are plenty of use cases for #244 where a given key might be used for any application with any padding mode, especially KMS/TPM-style applications where a given client might ask for a key operation involving any of them.

However, if you know ahead of time what the key's usage will be, the API in this PR is great.

tarcieri:
@lumag if you can rebase, the encryptor/decryptor types seem interesting

lumag:
Closing in favour of #259

Issue #253, opened 2023-01-19
PKCS#1 v1.5 signatures example on docs.rs doesn't work


rster2002:
I've been trying to get signing to work, but did not have any luck trying. One of my trouble shooting steps is to create a new rust binary package and try to run [this example](https://docs.rs/rsa/latest/rsa/#pkcs1-v15-signatures) but it doesn't seem to compile due to an unsatisfied trait bound.

main.rs:
```rust
use rsa::RsaPrivateKey;
use rsa::pkcs1v15::{SigningKey, VerifyingKey};
use rsa::signature::{Keypair, RandomizedSigner, SignatureEncoding, Verifier};
use sha2::{Digest, Sha256};

fn main() {
    let mut rng = rand::thread_rng();

    let bits = 2048;
    let private_key = RsaPrivateKey::new(&mut rng, bits).expect("failed to generate a key");
    let signing_key = SigningKey::<Sha256>::new_with_prefix(private_key);
    let verifying_key = signing_key.verifying_key();

    // Sign
    let data = b"hello world";
    let signature = signing_key.sign_with_rng(&mut rng, data);
    assert_ne!(signature.to_bytes().as_ref(), data.as_slice());

    // Verify
    verifying_key.verify(data, &signature).expect("failed to verify");
}
```

cargo.toml:
```toml
[package]
name = "rsa-test"
version = "0.1.0"
edition = "2021"

# See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html

[dependencies]
rand = "0.8.5"
rsa = "0.8.0"
sha2 = "0.10.6"
```

Environment:
* Windows 10: Version 10.0.19045 Build 19045
* `cargo --version`: cargo 1.65.0 (4bc8f24d3 2022-10-20)

Is there something I'm missing here or is the example on docs.rs outdated? Thanks in advance!

tarcieri:
Note that all of the docs on https://docs.rs are doctested in CI, so in general they will not ever be out-of-date for a given version.

The problem is you need the `oid` feature of the `sha2` crate enabled:

```
sha2 = { version = "0.10.6", features = ["oid"] }
```

We should probably add an explicit dependency on `sha2` with the `oid` feature to alleviate this problem.

rster2002:
I couldn't find that in the docs. Thanks for the help.

Issue #238, opened 2022-12-19
Confusion over `pkcs1v15::VerifyingKey::new` vs `::new_with_prefix`


tarcieri:
In #234 we had some people confused about signature verification failing with `::new`, and they had trouble discovering `::new_with_prefix`.

It would be good to either improve the documentation here, or eliminate the distinction between `::new` and `::new_with_prefix` and automatically determining if the prefix is needed (e.g. based on the OID of the provided digest).

It seems like other RSA libraries are able to make this determination automatically.

dignifiedquire:
I think it would be great if we had 

- `new()` -> determines it automatically
- `new_prefix(p: Prefix)` explicitly expects a prefix or not. With `enum Prefix { Yes, No }` (names can be bikeshedded) for when you want to enforce one or the other.

tarcieri:
I think part of the rationale for the current split is avoiding the need to have an `AssociatedOid` on the digest used.

Looking at [RFC8017 Appendix 2.4](https://www.rfc-editor.org/rfc/rfc8017#appendix-A.2.4), I'm unclear under what circumstances the OID can actually be omitted?

So perhaps the current `new` should be renamed to something like `new_unprefixed`, and `new_with_prefixed` should be renamed to `new`?

Issue #234, opened 2022-12-01
Interoperability issue with PKCS#1 v1.5 signature verification


cmcqueen:
I am trying to use this crate to verify a PKCS#1 v1.5 signature that was created with Python [pycryptodome](https://pypi.org/project/pycryptodome/).

The following Python + pycryptodome example code successfully verifies a PKCS#1 v1.5 signature for a message:

```python
from base64 import b64decode
from Crypto.Hash import SHA256
from Crypto.PublicKey import RSA
from Crypto.Signature import pkcs1_15 as pkcs1_15_sign

def main():
    public_key_pem = """-----BEGIN PUBLIC KEY-----
        MIIBIjANBgkqhkiG9w0BAQEFAAOCAQ8AMIIBCgKCAQEAzyaAz9Ak2ruY/9npoZ11
        S4SwZpw6Qx/4nk09xZh4nLsxEjXfXYZBzyLwLnEr040h6ZoGp4zOjAeBK+AsF2ev
        6hNLktwmF7xOXA0VtlTIXX3zl96Xax4ffeEy0uoyggGI9JvH+d+Zvb07OvAbn9js
        xEKVW2CUWHq9vHWUgB5mNf2B3iLj52NAjupiwyb7MfmnCCSWyjuJha428Vdsk4py
        iHn3k0Z1drI3kda9x21X5EfAkdsfm++FFJ5VpB7mRpGT3qg0zX9MtID8N0ZlTYCd
        7ZRlV2tPTOlOdH34tEUShESieKQBrUyxc0IT5wngxd3O6ZJ5CCLzGc4XSAeXkp8t
        9QIDAQAB
        -----END PUBLIC KEY-----"""
    signature64 = (
        "jdTNMwoSCf5bJWeCkKve5kOdKkuLtRokcItjLHEwGRcCnmmTChk6qmopKPdiGvJL"
        "w3iJfGHSh2Z71yERX35m/RVLvkbnnYQtssKFVDj49ttnDXkK/pcPLdbnWc2GPbLN"
        "v5XIpgiR2FPyp2PGbwYDoYNFnOI4ICJSBfKLjjYECCGbKKj4w5QBGzXyiCL0gV6l"
        "iS/7Wj1BCjwDb09FJF3oLGM5kpQlP3r1Ksh38cxsdUbBA/+29TOYclGEqw8VwgT9"
[...]

PR #255, opened 2023-01-20
Add `sha2` feature with `oid` subfeature enabled

tarcieri:
We seem to be running into a lot of people who are having trouble with PKCS#1 v1.5 signatures because the failure mode for the `oid` feature of the `sha2` crate being disabled is fairly unscrutable.

See #234, #253, and the semi-related tracking issue for #238.

If `rsa` has a `sha2` feature, we can always ensure `oid` is enabled, and this can be used in code examples. It also means users don't need two crates to create/verify PKCS#1 v1.5 signatures.

RSA is used commonly enough with the SHA2 family that this integration probably makes sense.

cc @lumag


dignifiedquire (approved):
sounds good

PR #258, opened 2023-01-30
Implementation of prime factors recovery

davxy:
Implements algorithm described in Appendix C of  https://nvlpubs.nist.gov/nistpubs/SpecialPublications/NIST.SP.800-56Br1.pdf

As a first draft, this is using the `rand_core::OsRng` to generate a random number in the range [2, n). 
Obviously this should be replaced somehow. One option is to pass an impl of a PRNG as a parameter, but maybe there is a better solution...

---

Given that:
> According to Fact 1 in [Boneh 1999], the probability that one of the values of y in an
> iteration of Step 3 reveals the factors of the modulus is at least 1/2, so on average, at most
> two iterations of that step will be required

and that:
- *gcd(g, n)  = 1*  (if not we immediately found a factor of n and we've finished...)
- *g^(ed - 1) = 1 (mod n)*  (e and d are valid RSA parameters and this property should hold)
- (e*d - 1) can be decomposed in odd and even factors: r * 2^t  (with r odd)

The only requirement to succeed is to pick a `g` such that:
1. g^(r·k) = 1 (for 1<= k <= 2^t)             <= this is always true when the assumptions above are true
2. g^(r·(k-1)) ≠ -1              

These two requirements are why the probability to succeed with a random `g` is 1/2.
That is, given a sequence of `g^z` (0 <= z <= ed -1) that for sure ends up with 1 at some point we will catch -1 or 1.

If 2 doesn't hold ... we'll try with another `g`.

As an attempt to remove the dependency on a random num generator given the super high probability to catch a correct `g`, maybe we can remove at all the generation of `g` via a proper random number generator and set it to something "good enough" (e.g `g = e · i`, with i =1..100) or something like that...

### DIFF
--- a/src/algorithms.rs
+++ b/src/algorithms.rs
@@ -250,7 +250,6 @@ pub fn recover_primes(n: &BigUint, e: &BigUint, d: &BigUint) -> Result<(BigUint,
                 let g = (g - &one) % n;
                 let p = n.gcd(&g);
                 let q = n / &p;
-                std::println!("count: {}", count);
                 return Ok((p, q));
             } else if g_next == n_min1 {
                 continue;
--- a/src/algorithms.rs
+++ b/src/algorithms.rs
@@ -207,12 +207,18 @@ fn inc_counter(counter: &mut [u8; 4]) {
     }
 }
 
-/// TODO
-/// Probabilistic algorithm that given `d` returns `p` and `q`
-pub fn recover_primes(n: &BigUint, e: &BigUint, d: &BigUint) -> Result<(BigUint, BigUint)> {
-    const ITER_LIMIT: usize = 100;
+/// Probabilistic algorithm to recover the prime factors of `n` given
+/// `d` and `e` components.
+///
+/// See Appendix C of [NIST-SP-800-56B-r1](https://nvlpubs.nist.gov/nistpubs/SpecialPublications/NIST.SP.800-56Br1.pdf)
+pub fn recover_primes<R: CryptoRngCore + ?Sized>(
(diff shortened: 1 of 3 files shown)


tarcieri:
> As a first draft, this is using the `rand_core::OsRng` to generate a random number in the range [2, n). 
Obviously this should be replaced somehow. One option is to pass an impl of a PRNG as a parameter, but maybe there is a better solution...

Replacing `rand_core::OsRng` with a PRNG seems good. I'd suggest parameterizing `recover_primes` with a `CryptoRngCore` so it's possible to plug in any RNG.

As for which one to use, `Hash_DRBG` is both FIPS approved and fairly parsimonious with the other dependencies of the crate. Unfortunately we don't have a ready-made implementation to use. Perhaps we could put one in e.g. the `digest` crate (or failing that, another crate, but it would be nice to not have to add an extra dependency just for this functionality)

tarcieri:
@davxy Appendix C.2 has a "Deterministic Prime-Factor Recovery" algorithm. Perhaps it would make sense to implement that one instead?

davxy:
Yeah, I think could be a good idea to have both.

I'm going to have a look at the deterministic algorithm before attempt an implementation, but not before the we :-D

tarcieri:
FYI, the deterministic algorithm isn't in SP 800-56B Rev 1.

It's only in Rev 2: https://nvlpubs.nist.gov/nistpubs/SpecialPublications/NIST.SP.800-56Br2.pdf

tarcieri:
Closing in favor of #380

PR #259, opened 2023-02-07
Add Encryption-related traits

lumag:
Follow the signing API and add a set of traits to be used for data encryption and decryption.


src/oaep.rs
```diff
@@ -174,48 +222,117 @@ impl fmt::Debug for Oaep {
 ///
 /// [PKCS#1 OAEP]: https://datatracker.ietf.org/doc/html/rfc8017#section-7.1
 #[inline]
-pub fn encrypt<R: CryptoRngCore, K: PublicKey>(
+fn encrypt<R: CryptoRngCore + ?Sized, K: PublicKey>(
```
tarcieri, line 225: For anyone else reviewing, this isn't a breaking change because the `oaep` module isn't currently public.

This PR makes it public though, and I approve of making these parts of the API surface private in that case.

src/traits.rs
```diff
@@ -0,0 +1,43 @@
...
+    /// Decrypt the given message.
+    fn decrypt_with_rng<R: CryptoRngCore + ?Sized>(
+        &self,
+        rng: &mut R,
+        ciphertext: &[u8],
+    ) -> Result<Vec<u8>>;
+}
+
+/// Encryption keypair with an associated encryption key.
+///
```
tarcieri, line 35: Nit: empty line

src/oaep.rs
lumag: Let's split this to a separate commit to ease review.

tarcieri (approved):
Looks good! I really like this API, and the traits that sort of mirror the `signature` crate

src/traits.rs
lumag: removed

PR #263, opened 2023-02-13
Fix pss sign/verify when key length is multiple of 8 + 1 bits.

dfabregat:
In the unusual case of having a key of length k*8 + 1 bits, `em_len` will be 1 byte shorter than `key_len`. The call to `raw_encryption_primitive` requires a buffer `key_len` bytes long.

Vice versa, we would be generating a one-byte too-long salt in `generate_salt`.

Finally, `emsa_pss_verify_pre` would fail by shifting 1 byte left by 8 bits.

### DIFF
--- a/src/pss.rs
+++ b/src/pss.rs
@@ -196,9 +196,10 @@ pub(crate) fn verify<PK: PublicKey>(
 
     let em_bits = pub_key.n().bits() - 1;
     let em_len = (em_bits + 7) / 8;
-    let mut em = pub_key.raw_encryption_primitive(sig, em_len)?;
+    let key_len = pub_key.size();
+    let mut em = pub_key.raw_encryption_primitive(sig, key_len)?;
 
-    emsa_pss_verify(hashed, &mut em, em_bits, None, digest)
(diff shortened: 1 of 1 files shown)


dignifiedquire:
is there an actual use case for keys of that size, or should we just restrict the sizes?

dfabregat:
I personally do not remember encountering any special use case of this type. I've generally seen multiples of 512. On the other hand, OpenSSL, for instance, seems to support all key lengths (within whichever range they allow).

In the end, it is a design decision. I personally would rather support this case, but it is just my opinion. Alternatively, one can restrict and see if anyone ever raises a request for this. Then, we learn in which context would one have such a need :)

dignifiedquire:
@tarcieri any thoughts on this?

dignifiedquire (approved):
LGTM, thanks

dignifiedquire:
@tarcieri agreed, merging this, as this in all cases improves the status quo 

thanks for catching and fixing @dfabregat

dfabregat:
Always glad to contribute. And thank you guys for your feedback :smiley:

Issue #261, opened 2023-02-09
Should `PaddingScheme` and `SignatureScheme` be object safe?


tarcieri:
In #260 it was pointed out that `PaddingScheme` isn't object safe.

I think the purpose of these traits is to provide runtime dynamism around key usages, which makes the lack of object safety a bit problematic.

Especially with #259 complementing the existing signature-related `SigningKey`/`VerifyingKey` types with nice monomorphic purpose-dedicated encryption-related counterparts, that would make for a nice split between those providing a performance-oriented, type-safe monomorphic API, and `PaddingScheme`/`SignatureScheme` could evolve to focus on runtime dynamism.

Making that work would involve quite a bit of refactoring, getting rid of the generic parameters on the methods and replacing them with `dyn Trait`, namely for the `PrivateKey`, `PublicKey`, and `CryptoRngCore` traits. This would in turn require finding some way to make the `PrivateKey` and `PublicKey` traits object safe. At that point, we might consider trying to recombine the four traits (`PaddingScheme`, `SignatureScheme`, `PublicKey`, `PrivateKey`) somehow.

tarcieri:
One way I could see a refactor working is reversing the relationship between the `*Key` and `*Scheme` traits so that the `*Key` traits accept the `*Scheme` traits as arguments.

This would make the `*Scheme` traits only responsible for constructing padded messages that the `*Key` traits then act on.  That would make it a lot easier for the `*Scheme` traits to be object safe, and also minimize their responsibilities.

tarcieri:
I experimented with this in #310, however all methods of these traits need to move out of `self`, which means they can't be object safe.

So closing as this won't work, unless we redesign the traits.

Issue #99, opened 2021-06-21
Checks performed in check_public seem insufficient


randombit:
The `check_public` function nominally checks "that the public key is well formed and has an exponent within acceptable bounds."
https://github.com/RustCrypto/RSA/blob/master/src/key.rs#L642

But it also accepts even exponents, as long as they are within the defined range.

And it does not perform any checks on the modulus, accepting values including zero, even numbers, or modulus smaller than the public exponent.

Is accepting such keys intentional?

dignifiedquire:
no, this is likely a carry over from the fact thah originally it was not possible to specify a custom exponent. So this should be fixed

dfabregat:
In fact, the checks should be carried out in a couple of places. For instance, passing an even exponent to the function `RsaPrivateKey::new_with_exp` will hang forever, since the computation of the modular inverse in src/algorithms.rs:128 will never succeed. This simple test showcases the issue:

```
    #[test]
    fn test_new_with_exp() {
        let mut rng = rand::thread_rng();
        let bits = 2048;
        let _priv_key = RsaPrivateKey::new_with_exp(&mut rng, bits, &BigUint::from_u64(65536).unwrap());
    }  
```

I think this should be given some thought to make sure checks are in place where needed and not much duplicated code is introduced. Maybe one could also set a lower bound to the modulus size (e.g., 1024), and add to the checks. NIST SP 800-56B Revision 2 (section 6.4) may provide some guidance.

If you give me some time to get familiarized with the code, I can propose some changes.

tarcieri:
We added an initial set of public key checks in #170 and #176.

There currently aren't any such checks on private keys, beyond checking that there are at least two primes.

Checks on both public and private keys could definitely be further improved. A minimum modulus size definitely seems like a good idea (with the ability to bypass it through an `_unchecked` API, which seems important to @dignifiedquire's goals of supporting exotic legacy OpenPGP keys).

I think the private key APIs currently permit a number of potential misuses (see above-mentioned exponent-related DoS) which it would be good to mitigate.

dfabregat:
Thanks for your feedback @tarcieri. I'd be happy to know more about those exotic/legacy use cases. @dignifiedquire: Any pointer where this has been discussed or documented?

dignifiedquire:
[...]

PR #271, opened 2023-03-06
v0.9.0-pre.0

tarcieri:


dignifiedquire (approved):
I am fine with the changes in general, but as mentioned in the other issue going up to 1.65 is unfortunately somewhat problematic for downstream deps.

tarcieri:
@dignifiedquire I'm not in a hurry to get a release out so long as there's a prerelease.

IMO it'd be good to address the issues raised in #99 as well as #238 and #261.

Perhaps in the intervening time the blockers on 1.65 upgrades will be addressed?

tarcieri:
@dignifiedquire another option would be to make `pkcs1`/`pkcs8` optional dependencies (although that probably doesn't help if you're using them)

Issue #273, opened 2023-03-10
Example in ReadMe.md is not update to date


brmmm3:
In older versions of rsa trait the implementation like in the ReadMe example worked. But with the newer version it doesn't compile with the following errors:
```
add `dyn` keyword before this trait
   |
16 |     let dec_data = priv_key.decrypt(<dyn PaddingScheme>::new_pkcs1v15_encrypt(), &enc_data).expect("failed to decrypt");
```

and

```
error[E0599]: no function or associated item named `new_pkcs1v15_encrypt` found for trait object `dyn PaddingScheme` in the current scope
  --> tests/pkcs1v15.rs:16:52
```

Please update the example.

tarcieri:
It'd probably be good to shift to using `#![doc = include_str!("../README.md")]` in lib.rs, which both unifies the documentation in the README.md and rustdoc, and also doctests the examples in README.md so as to avoid this problem in perpetuity.

tarcieri:
Fixed in #276

PR #276, opened 2023-03-22
Updating README example

Ricardicus:
I wanted to test this crate using the example in the readme, but it failed, so
here in an updated README so that it builds.
I also included the README example among the doctests, so that it keeps up in the future.

I did find updated documentation on https://docs.rs/rsa/latest/rsa/ so it seems just this file
has been out of date.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -1,3 +1,4 @@
+#![doc = include_str!("../README.md")]
 //! RSA Implementation in pure Rust.
 //!
 //! It supports several schemes described in [RFC8017]:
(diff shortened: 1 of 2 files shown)


src/lib.rs
```diff
@@ -209,6 +209,10 @@
 #![doc(html_logo_url = "https://raw.githubusercontent.com/RustCrypto/meta/master/logo_small.png")]
 #![warn(missing_docs)]
 
+#[doc = include_str!("../README.md")]
```
tarcieri, line 212: As I mentioned in #273, I think it'd be good to switch to `#![doc = include_str!("../README.md")]` and include the README verbatim at the top of the rustdoc similar to how we do with many of the other @RustCrypto crates

tarcieri (approved):

PR #277, opened 2023-03-27
pss: add function to get salt length from RSA PSS keys

lumag:
To create proper RsaPssParams it is necessary to determine the salt length specified for the signing key. Add a function to get it from signing keys.


tarcieri (approved):

tarcieri:
Looks good, thanks!

PR #278, opened 2023-03-27
AssociatedAlgorithmIdentifier implementation

lumag:
This is an implementation of a trait defined at https://github.com/RustCrypto/formats/pull/954

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -24,10 +24,13 @@ subtle = { version = "2.1.1", default-features = false }
 digest = { version = "0.10.5", default-features = false, features = ["alloc", "oid"] }
 pkcs1 = { version = "0.7.2", default-features = false, features = ["alloc", "pkcs8"] }
 pkcs8 = { version = "0.10.2", default-features = false, features = ["alloc"] }
-serde = { version = "1.0.103", optional = true, default-features = false, features = ["derive"] }
-sha2 = { version = "0.10.6", optional = true, default-features = false, features = ["oid"] }
 signature = { version = "2", default-features = false , features = ["digest", "rand_core"] }
 zeroize = { version = "1", features = ["alloc"] }
+
+# optional dependencies
+serde = { version = "1.0.103", optional = true, default-features = false, features = ["derive"] }
+sha1 = { version = "0.10.5", optional = true, default-features = false, features = ["oid"] }
+sha2 = { version = "0.10.6", optional = true, default-features = false, features = ["oid"] }
 
 [dev-dependencies]
 base64ct = { version = "1", features = ["alloc"] }
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -19,10 +19,11 @@ num-integer = { version = "0.1.39", default-features = false }
 num-iter = { version = "0.1.37", default-features = false }
 rand_core = { version = "0.6.4", default-features = false }
 byteorder = { version = "1.3.1", default-features = false }
+const-oid = { version = "0.9", default-features = false }
 subtle = { version = "2.1.1", default-features = false }
 digest = { version = "0.10.5", default-features = false, features = ["alloc", "oid"] }
-pkcs1 = { version = "0.7.1", default-features = false, features = ["alloc", "pkcs8"] }
-pkcs8 = { version = "0.10", default-features = false, features = ["alloc"] }
+pkcs1 = { version = "0.7.2", default-features = false, features = ["alloc", "pkcs8"] }
+pkcs8 = { version = "0.10.2", default-features = false, features = ["alloc"] }
 serde = { version = "1.0.103", optional = true, default-features = false, features = ["derive"] }
 sha2 = { version = "0.10.6", optional = true, default-features = false, features = ["oid"] }
 signature = { version = "2", default-features = false , features = ["digest", "rand_core"] }
--- a/src/pkcs1v15.rs
+++ b/src/pkcs1v15.rs
@@ -12,7 +12,10 @@ use core::fmt::{Debug, Display, Formatter, LowerHex, UpperHex};
 use core::marker::PhantomData;
 use digest::Digest;
 use pkcs8::{
-    spki::{der::AnyRef, AlgorithmIdentifierRef, AssociatedAlgorithmIdentifier},
+    spki::{
+        der::AnyRef, AlgorithmIdentifierRef, AssociatedAlgorithmIdentifier,
+        SignatureAlgorithmIdentifier,
+    },
     AssociatedOid, Document, EncodePrivateKey, EncodePublicKey, SecretDocument,
 };
 use rand_core::CryptoRngCore;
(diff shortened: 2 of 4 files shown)


rozbb:
Thanks! Could you possibly 1) make the ID string consts at the top of the file, and 2) put some links in the comments for where the strings come from?

Cargo.toml
```diff
@@ -52,9 +57,10 @@ std = ["digest/std", "pkcs1/std", "pkcs8/std", "rand_core/std", "signature/std"]
 pem = ["pkcs1/pem", "pkcs8/pem"]
 pkcs5 = ["pkcs8/encryption"]
 getrandom = ["rand_core/getrandom"]
+algoid = ["spki/alloc", "const-oid"]
```
tarcieri, line 60: I know it's nauseatingly verbose but I think I'd probably prefer `algorithm-identifier` or failing that `algorithm-id` for this just as there aren't otherwise canonical abbreviations

src/pkcs1v15.rs
```diff
@@ -436,6 +436,26 @@ where
     }
 }
 
+#[cfg(all(feature = "algoid", feature = "sha1"))]
+impl spki::DynAssociatedAlgorithmIdentifier for SigningKey<sha1::Sha1> {
+    fn algorithm_identifier(&self) -> spki::AlgorithmIdentifierOwned {
+        spki::AlgorithmIdentifierOwned {
+            oid: const_oid::ObjectIdentifier::new_unwrap("1.2.840.113549.1.1.5"),
+            parameters: None,
```
baloo, line 444: All the parameters needs to be set explicitly to NULL as per https://www.rfc-editor.org/rfc/rfc4055#section-1.2

baloo: (This is specific to x509 here, but I don't think it would hurt?)

lumag: Done

src/pkcs1v15.rs
```diff
@@ -436,6 +439,71 @@ where
...
+    const ALGORITHM_IDENTIFIER: AlgorithmIdentifierRef<'static> = pkcs1::ALGORITHM_ID;
+}
+
+#[cfg(all(feature = "algorithm-identifier", feature = "sha1"))]
+impl spki::SignatureAlgorithmIdentifier for SigningKey<sha1::Sha1> {
+    type Params = AnyRef<'static>;
+
+    const SIGNATURE_ALGORITHM_IDENTIFIER: AlgorithmIdentifierRef<'static> =
+        AlgorithmIdentifierRef {
+            oid: const_oid::ObjectIdentifier::new_unwrap("1.2.840.113549.1.1.5"),
```
lumag, line 458: @tarcieri @dignifiedquire 
Actually I'm thinking about a special `trait RsaEncryptionOid` which then can be implemented for `Digest`.
The `RsaEncryptionOid` can go into traits (Ugh), and then digest implementations can implenent them on their own.

WDYT?

tarcieri: In `ecdsa` I'm mapping from the digest OIDs to the ECDSA OIDs, in a manner which could work as a `const fn` potentially. Would something like that work here?

lumag: I thought about that, but strictly speaking I don't like the idea of handling all the mapping here. Initially I had just sha1/sha256 (which was bearable). Then it got extended to all SHA2 functions. And there exist MD4/MD5/MD2/etc. This is unlike ECDSA where there are just 4 digest functions.

lumag: I think I'm going to split the PSS common PKCS1v15 implementation (so that it can be merged now) and leave pkcs1v15 Signature stuff as a draft.

tarcieri: Alternatively you could define the carrier trait in the RSA crate and conditionally implement it when the e.g. `sha2` feature is enabled.

I wouldn't worry too much about extended digests that aren't commonly used, at least at first, particularly broken ones like MD2, MD4, and MD5. We can cross that bridge when we get there.

lumag: Well Sha1 is broken too. It's all about the legacy and existing files. The bonus of the global `RsaSignatureOid` is that we won't have to bloat the RSA crate with tons of features.

tarcieri: It already has a `sha2` feature, FWIW.

The features make thing nicer for downstream users. It's one less dependency for them to import, and it took care of all of the confusion around enabling the `oid` feature of `sha2`.

lumag: We can reuse the "oid" ;-) I fear that selecting RSA's features might also become an issue.

lumag: I have implemented the `SignatureAlgorithmIdentifier` for PKCS1v15 keys in a way that doesn't require `panic!`, provide compile time checks and can be migrated to generic traits/in-digest-trait implementation.

lumag:
@baloo this PR has support for PKCS1v15. Could you please check whether it works for you?

baloo:
Yeah, I was missing the blanket impl for `SignatureAlgorithmIdentifier`. That works for me.

src/pkcs1v15.rs
```diff
@@ -774,6 +824,46 @@ impl EncryptingKeypair for DecryptingKey {
...
+    #[cfg(feature = "sha1")]
+    impl RsaSignatureAssociatedOid for sha1::Sha1 {
+        const OID: ObjectIdentifier =
+            const_oid::ObjectIdentifier::new_unwrap("1.2.840.113549.1.1.5");
+    }
+
+    #[cfg(feature = "sha2")]
+    impl RsaSignatureAssociatedOid for sha2::Sha224 {
+        const OID: ObjectIdentifier =
+            const_oid::ObjectIdentifier::new_unwrap("1.2.840.113549.1.1.14");
```
baloo, line 845:
```suggestion
            const_oid::db::rfc5912::SHA_224_WITH_RSA_ENCRYPTION
```

lumag: Yes, this was bugging me too. But I think the overall design now is not to add a dependency on `const-oid/db`.
@tarcieri ?

lumag:
> Which of these (https://github.com/RustCrypto/RSA/pull/278 vs https://github.com/RustCrypto/RSA/pull/284) should I be reviewing? They have many of the same changes.

This ons (#278). #284 was a spin-off w/o pkcs1v15 changes which were controvercial.

src/pss.rs
```diff
@@ -689,6 +698,61 @@ where
...
+{
+    const ID_MGF_1: ObjectIdentifier = ObjectIdentifier::new_unwrap("1.2.840.113549.1.1.8");
+    const ID_RSASSA_PSS: ObjectIdentifier = ObjectIdentifier::new_unwrap("1.2.840.113549.1.1.10");
+
+    let salt_len = salt_len.map_or(RsaPssParams::SALT_LEN_DEFAULT, |l| l as u8);
+
+    /*
+     * We do not expect that any of this functions fails, unless the library is broken, so it
+     * is safe to use unwrap()
+     */
```
tarcieri, line 713: Perhaps use `expect` instead of `unwrap`?

lumag:
@tarcieri `rsa` crate doesn't have direct dependency on `spki`/`der`. Would it be possible to cut a dummy point release for `pkcs8` to bump its internaly dependecy? Or we'd better declare `rsa` -> `spki` dependency?

tarcieri:
@lumag already in progress: https://github.com/RustCrypto/formats/pull/983

lumag:
Thanks! didn't notice it.

lumag:
Ugh. @tarcieri Another request, but now for pkcs1, to get `RsaPssParams::SALT_LEN_DEFAULT`

tarcieri (approved):

lumag:
@tarcieri thank you!

PR #284, opened 2023-04-02
feat: implement associated AlgorithmIdentifier traits for PKCS1v15 an…

lumag:
…d PSS keys

Allow getting the algorithm identifiers for signing keys

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -19,6 +19,7 @@ num-integer = { version = "0.1.39", default-features = false }
 num-iter = { version = "0.1.37", default-features = false }
 rand_core = { version = "0.6.4", default-features = false }
 byteorder = { version = "1.3.1", default-features = false }
+const-oid = { version = "0.9", default-features = false }
 subtle = { version = "2.1.1", default-features = false }
 digest = { version = "0.10.5", default-features = false, features = ["alloc", "oid"] }
 pkcs1 = { version = "0.7.1", default-features = false, features = ["alloc", "pkcs8"] }
@@ -59,3 +60,8 @@ rustdoc-args = ["--cfg", "docsrs"]
 
 [profile.dev]
 opt-level = 2
+
+[patch.crates-io]
+pkcs1 = { git = "https://github.com/RustCrypto/formats" }
+pkcs8 = { git = "https://github.com/RustCrypto/formats" }
+spki = { git = "https://github.com/RustCrypto/formats" }
--- a/src/pss.rs
+++ b/src/pss.rs
@@ -12,10 +12,19 @@
 use alloc::boxed::Box;
 use alloc::vec::Vec;
 use core::fmt::{self, Debug, Display, Formatter, LowerHex, UpperHex};
-
 use core::marker::PhantomData;
+
+use const_oid::{AssociatedOid, ObjectIdentifier};
 use digest::{Digest, DynDigest, FixedOutputReset};
-use pkcs8::{Document, EncodePrivateKey, EncodePublicKey, SecretDocument};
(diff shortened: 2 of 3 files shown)


Cargo.toml
```diff
@@ -23,11 +23,15 @@ subtle = { version = "2.1.1", default-features = false }
 digest = { version = "0.10.5", default-features = false, features = ["alloc", "oid"] }
 pkcs1 = { version = "0.7.1", default-features = false, features = ["alloc", "pkcs8"] }
 pkcs8 = { version = "0.10", default-features = false, features = ["alloc"] }
-serde = { version = "1.0.103", optional = true, default-features = false, features = ["derive"] }
-sha2 = { version = "0.10.6", optional = true, default-features = false, features = ["oid"] }
 signature = { version = "2", default-features = false , features = ["digest", "rand_core"] }
 zeroize = { version = "1", features = ["alloc"] }
 
+# optional dependencies
+const-oid = { version = "0.9", optional = true }
```
tarcieri, line 30: This is already a transitive dependency of `pkcs1` and `pkcs8`, which are both hard dependencies. You might as well make it a hard dependency.

Cargo.toml
```diff
@@ -23,11 +23,15 @@ subtle = { version = "2.1.1", default-features = false }
...
+serde = { version = "1.0.103", optional = true, default-features = false, features = ["derive"] }
+sha2 = { version = "0.10.6", optional = true, default-features = false, features = ["oid"] }
+spki = { version = "0.7.0", optional = true, default-features = false }
```
tarcieri, line 33: This is a transitive dependency of `pkcs8`, which is a hard dependency.

Cargo.toml
lumag: done

Cargo.toml
lumag: done

lumag:
@tarcieri 
Note, I followed the `ecdsa` example, this PR contains `patch.crates.io`.  It should be removed after releasing spki.

tarcieri:
@lumag happy to do that if @baloo is happy with the API

baloo:
I'm pretty happy with the API. I was missing support on PKCS#1 v1.5, see https://github.com/lumag/RSA/pull/3

Cargo.toml
```diff
@@ -52,10 +53,16 @@ std = ["digest/std", "pkcs1/std", "pkcs8/std", "rand_core/std", "signature/std"]
 pem = ["pkcs1/pem", "pkcs8/pem"]
 pkcs5 = ["pkcs8/encryption"]
 getrandom = ["rand_core/getrandom"]
+algorithm-identifier = ["pkcs1/pkcs8"]
```
tarcieri, line 56: This is already enabled by default? I don't see any need to feature-gate `algorithm-identifier` whatsoever

lumag: done

src/pss.rs
```diff
@@ -689,6 +698,61 @@ where
...
+{
+    const ID_MGF_1: ObjectIdentifier = ObjectIdentifier::new_unwrap("1.2.840.113549.1.1.8");
+    const ID_RSASSA_PSS: ObjectIdentifier = ObjectIdentifier::new_unwrap("1.2.840.113549.1.1.10");
+
+    let salt_len = salt_len.map_or(RsaPssParams::SALT_LEN_DEFAULT, |l| l as u8);
+
+    /*
+     * We do not expect that any of this functions fails, unless the library is broken, so it
+     * is safe to use unwrap()
+     */
```
tarcieri, line 713: Perhaps it's better to use `expect` here?

PR #290, opened 2023-04-10
pkcs1v15: make `*_with_prefix` methods the default

### LINKED ISSUE #238: Confusion over `pkcs1v15::VerifyingKey::new` vs `::new_with_prefix`

tarcieri:
Renames the following:

- `SigningKey::new` => `SigningKey::new_unprefixed`
- `SigningKey::new_with_prefix` => `SigningKey::new`
- `VerifyingKey::new` => `VerifyingKey::new_unprefixed`
- `VerifyingKey::new_with_prefix` => `VerifyingKey::new`

The `*_with_prefix` methods are preserved with a deprecation warning, which should help people migrate to the new versions.

Closes #238

### DIFF
--- a/src/pkcs1v15.rs
+++ b/src/pkcs1v15.rs
@@ -423,8 +423,8 @@ where
     ///
     /// ## Note: unprefixed signatures are uncommon
     ///
-    /// In most cases you'll want to use [`SigningKey::new_with_prefix`].
-    pub fn new(key: RsaPrivateKey) -> Self {
+    /// In most cases you'll want to use [`SigningKey::new`].
+    pub fn new_unprefixed(key: RsaPrivateKey) -> Self {
         Self {
             inner: key,
             prefix: Vec::new(),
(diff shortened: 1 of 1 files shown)


tarcieri:
Per #238 having an empty prefix seems to be quite rare.

People are reaching for `new` expecting it to work, because in most other environments prefixes seem to "just work" by default.

This change might be confusing for anyone who happens to actually be using empty prefixes, though I haven't encountered anyone who is personally.

tarcieri:
cc @lumag

lumag:
Ack, LGTM.

dignifiedquire (approved):
👍

PR #291, opened 2023-04-14
pss: make `*_with_salt` methods the default

tarcieri:
Similar to #290, this renames the following:

- `Pss::new` => `Pss::new_unsalted`
- `Pss::new_with_salt` => `Pss:new`
- `Pss::new_blinded` => `Pss::new_blinded_unsalted`
- `Pss::new_blinded_with_salt` => `Pss::new_blinded`
- `SigningKey::new` => `SigningKey::new_unsalted`
- `SigningKey::new_with_salt_len` => `SigningKey::new`
- `SigningKey::random` => `SigningKey::random_unsalted`
- `SigningKey::random_with_salt_len` => `SigningKey::random`

The `*_with_salt` methods are preserved with a deprecation warning, which should help people migrate to the new versions.

This also removes the `From<RsaPrivateKey>` impl for `SigningKey`, since users should consider up front whether or not they need a salt rather than defaulting to unsalted.

cc @lumag

### DIFF
--- a/src/pss.rs
+++ b/src/pss.rs
@@ -51,8 +51,17 @@ pub struct Pss {
 }
 
 impl Pss {
+    /// New PSS padding for the given digest with a salt value of the given length.
+    pub fn new<T: 'static + Digest + DynDigest + Send + Sync>(salt_len: usize) -> Self {
+        Self {
+            blinded: false,
+            digest: Box::new(T::new()),
+            salt_len: Some(salt_len),
+        }
+    }
+
     /// New PSS padding for the given digest.
-    pub fn new<T: 'static + Digest + DynDigest + Send + Sync>() -> Self {
+    pub fn new_unsalted<T: 'static + Digest + DynDigest + Send + Sync>() -> Self {
         Self {
             blinded: false,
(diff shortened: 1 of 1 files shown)


lumag:
Generic comment: do we need the without salt functions at all? Let me suggest another approach.

tarcieri:
I don't actually know if anyone uses unsalted PSS.

Indeed It seems weird to have a separate API for it when someone who really wants it can specify a salt length of 0.

Perhaps we should just get rid of the "unsalted" functions to simplify the API?

lumag:
Yes, that's my idea. Will post it shortly.

tarcieri:
I can just remove the `_unsalted` methods from this PR?

lumag:
@tarcieri slightly more than that (especially on the verification side).

lumag:
@tarcieri done, https://github.com/RustCrypto/RSA/pull/294

tarcieri:
I like #294 as an alternative to this, but I'll leave this open for now until people have a chance to discuss

tarcieri:
Closed in favor of #294

PR #292, opened 2023-04-15
lib: stop using deprecated SigningKey<D>::new_with_prefix

lumag:


tarcieri:
Huh, surprised that didn't trigger a deprecation warning

lumag:
It does, but only when doing `cargo test`. And in that case it is captured by default. I passed `--nocapture` when debugging a test and thus noticed this issue.

tarcieri (approved):

PR #293, opened 2023-04-15
Rename `Pkcs1v15Sign::new_raw` to `Pkcs1v15Sign::new_unprefixed`

tarcieri:
Following #290, which amended `pkcs1v15::SigningKey`, this commit makes a corresponding change to `Pkcs1v15Sign` so the method name is consistent with `SigningKey::new_unprefixed`.

cc @lumag

### DIFF
--- a/src/pkcs1v15.rs
+++ b/src/pkcs1v15.rs
@@ -83,14 +83,22 @@ impl Pkcs1v15Sign {
         }
     }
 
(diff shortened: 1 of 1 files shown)


lumag:
LGTM (please note that I don't use padding schemas).
Maybe @dignifiedquire does?

PR #294, opened 2023-04-15
Remove 'unsalted' PSS handling

lumag:
The 'unsalted' PSS handling is not behaving like one would expect. It doesn't use salt of 0 length, but insteaad it uses some  defaults which are not properly documented. Make the `salt_len` mandatatory for both singing and verification.

### DIFF
--- a/src/pss.rs
+++ b/src/pss.rs
@@ -103,7 +103,7 @@ impl SignatureScheme for Pss {
     }
 
     fn verify<Pub: PublicKey>(mut self, pub_key: &Pub, hashed: &[u8], sig: &[u8]) -> Result<()> {
-        verify(pub_key, hashed, sig, &mut *self.digest)
+        verify(pub_key, hashed, sig, &mut *self.digest, self.salt_len)
     }
 }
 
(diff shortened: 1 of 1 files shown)


tarcieri (approved):
Looks good to me!

I don't understand the potential real-world usages of unsalted PSS, but from a theoretical perspective it makes no sense: the purpose of having a salt is to increase the tightness of the security proof. I don't understand why one would want to use PSS in a way which defeats the security proof.

So that said, I'm all for removing the unsalted API.

I think we could further improve the API by having one where the default salt length equals the output length of the digest function, but I'm happy to open a followup PR to add that.

lumag:
@tarcieri sounds good to me.

cc @roblabla @dignifiedquire

tarcieri:
@lumag alternatively, you could implement ^^^ API in this PR, which would avoid the need to deprecate anything.

Instead, change `new` to be salted, but infer the salt size from the digest function.

lumag:
Sure, I'll do that in a minute

lumag:
done

tarcieri (approved):

PR #297, opened 2023-04-17
Impl `Signer` for `pss::SigningKey`

tarcieri:
Gated under the `getrandom` feature.

Calls `RandomizedSigner` with `OsRng`.

cc @lumag


lumag:
Don't we need to have `getrandom` declared as a feature in Cargo.toml?

tarcieri:
@lumag it already is

lumag:
I see :-)

Issue #272, opened 2023-03-09
RSA signatures larger than the modulus are accepted


randombit:
`raw_encryption_primitive` is used for verifying signatures. It does not check that the decoded integer value is less than the modulus, thus if `s` is a valid signature it will also accept `s+k*n` where `n` is the public modulus and `k` is any positive integer.

This introduces signature malleability, which is probably not an enormous problem in most applications, but neither does it seem desirable.

A related issue is that signatures are decoded as big integers but it is never checked that the length of the signature is equal to the public modulus length. So if `sig` is the binary encoding of a valid signature, prefixing that signature with any number of zero bytes will also be accepted as valid.

This may affect ciphertext decryption as well, but I haven't checked this.

The following patch to the tests demonstrates the issues

```
diff --git a/src/pkcs1v15.rs b/src/pkcs1v15.rs
index 5d9bac4..50b6dc1 100644
--- a/src/pkcs1v15.rs
+++ b/src/pkcs1v15.rs
@@ -1152,15 +1152,23 @@ mod tests {
             (
                 "Test.\n",
                 hex!(
-                    "a4f3fa6ea93bcdd0c57be020c1193ecbfd6f200a3d95c409769b029578fa0e33"
+                    "00a4f3fa6ea93bcdd0c57be020c1193ecbfd6f200a3d95c409769b029578fa0e33"
                     "6ad9a347600e40d3ae823b8c7e6bad88cc07c1d54c3a1523cbbb6d58efc362ae"
                 ),
                 true,
             ),
             (
                 "Test.\n",
+                // signature + modulus
                 hex!(
-                    "a4f3fa6ea93bcdd0c57be020c1193ecbfd6f200a3d95c409769b029578fa0e33"
+                    "01578d09b86db9c85d997c8e8b0e34c9076782843561884f09b2969a0e83d8a88033052e71e182beb26f0d9e2363a5b64bfd7040c7a88609b7375d2e3ef3e1ffc3"
+                ),
+                true,
+            ),
+            (
+                "Test.\n",
+                hex!(
+                    "00a4f3fa6ea93bcdd0c57be020c1193ecbfd6f200a3d95c409769b029578fa0e33"
                     "6ad9a347600e40d3ae823b8c7e6bad88cc07c1d54c3a1523cbbb6d58efc362af"
                 ),
                 false,
```

tarcieri:
[...]

Issue #220, opened 2022-11-07
Use `BigUint` for the Signature internal representations


lumag:
Ideally we could replace the internal representation of a `Signature`s with a `BigUint` now, which would eliminate the need for an intermediate `Box<[u8]>` when signing/verifying.

_Originally posted by @tarcieri in https://github.com/RustCrypto/RSA/pull/217#discussion_r1009872868_

tarcieri:
cc @npmccallum 

There are many APIs that currently take a slice which would need at least an equivalent API which operates on a `BigUint`.

I think we could potentially consider removing some of the old APIs from the public API as well.

tarcieri:
Implemented in #298 and #300

PR #298, opened 2023-04-18
pkcs1v15: use `BigUint` as `Signature`'s inner type

tarcieri:
This one half of #220.

Doing anything with a signature involves converting it from bytes into a `BigUint`, so this changes the inner type the latter which is more useful.

It should also help address #272, since it will enable doing those sort of checks more eagerly.

cc @lumag

### DIFF
--- a/src/raw.rs
+++ b/src/raw.rs
@@ -1,7 +1,7 @@
 use alloc::vec::Vec;
 use num_bigint::BigUint;
 use rand_core::CryptoRngCore;
-use zeroize::Zeroize;
+use zeroize::Zeroizing;
 
 use crate::errors::Result;
 use crate::internals;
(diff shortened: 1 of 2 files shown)


src/raw.rs
```diff
@@ -1,15 +1,21 @@
...
 pub trait EncryptionPrimitive {
     /// Do NOT use directly! Only for implementors.
-    fn raw_encryption_primitive(&self, plaintext: &[u8], pad_size: usize) -> Result<Vec<u8>>;
+    fn raw_encryption_primitive(&self, plaintext: &[u8], pad_size: usize) -> Result<Vec<u8>> {
+        let int = Zeroizing::new(BigUint::from_bytes_be(plaintext));
+        self.raw_int_encryption_primitive(&int, pad_size)
+    }
+
+    fn raw_int_encryption_primitive(&self, plaintext: &BigUint, pad_size: usize)
+        -> Result<Vec<u8>>;
```
tarcieri, line 18: I'm not super enthusiastic about this API change, but I did it this way to reduce overall churn (i.e. so I don't have to change all of the existing usages of `raw_encryption_primitive`/`raw_decryption_primitive`.

I'm also not sure what these traits are for... are they intended to be implemented by e.g. hardware accelerators?

lumag: I think it might be better to drop these traits. Hardware accelerators will most probably implement the `src/key.rs` API rather than `src/raw.rs`.

lumag: This change looks good to me. Let's land it, I'll post the trait removal on top of it.

tarcieri: Cool, can cut a prerelease after that

PR #299, opened 2023-04-18
feat: switch to RsaPssParams::new()

lumag:
Switch to using RsaPssParams::new() from pkcs1 crate. This fixes the issue reported by zlint for x509-cert, where SHA* digest algorithms had empty parameters instead of NULL parameters (as required by Mozilla policy and permitted by RFC4055).

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -268,9 +268,9 @@ dependencies = [
 
 [[package]]
 name = "pkcs1"
-version = "0.7.2"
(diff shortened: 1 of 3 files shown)


lumag:
cc @baloo @dignifiedquire

tarcieri (approved):

lumag:
@tarcieri thank you! Could you please cut another prerelease?

PR #301, opened 2023-04-18
feat: make the implementation generic around the PublicKey/PrivateKey

lumag:
Use generic traits `PublicKey` / `PrivateKey` as much as possible instead of using the exact Rsa types.


lumag:
This is just a preview of another direction in comparison with #300 .

tarcieri:
I very much like #300 over this one, FWIW

PR #304, opened 2023-04-19
Continue refactoring of RSA crate internals

lumag:
Further refactoring of the RSA crate. Clean up internal interfaces, split software RSA implementation from the wrapping code.


lumag:
cc @dignifiedquire @tarcieri @str4d

src/algorithms/rsa.rs
```diff
@@ -0,0 +1,179 @@
...
+use num_traits::{One, Signed, Zero};
+use rand_core::CryptoRngCore;
+use zeroize::Zeroize;
+
+use crate::errors::{Error, Result};
+use crate::keytraits::{PrivateKeyParts, PublicKeyParts};
+
+/// Raw RSA encryption of m with the public key. No padding is performed.
+#[inline]
+pub fn rsa_encrypt<K: PublicKeyParts>(key: &K, m: &BigUint) -> Result<BigUint> {
```
tarcieri, line 15: This feels like a hazmat thing, and should probably carry a warning about the dangers of misused unpadded RSA

lumag: We can not put it under hazmat as it is used internally. Please note that that all algorithms modules are `pub(crate)` at best. But yes, the warning is probably a good idea.

src/lib.rs
```diff
@@ -222,8 +222,9 @@ pub use num_bigint::BigUint;
 pub use rand_core;
 pub use signature;
 
-pub mod algorithms;
+mod algorithms;
```
tarcieri, line 225: Aah, ok. Maybe make the methods within explicitly `pub(crate)` for clarity?

src/algorithms/rsa.rs
tarcieri: Now I see: https://github.com/RustCrypto/RSA/pull/304#pullrequestreview-1392921745

src/lib.rs
lumag: Yep. Sounds like a good idea. I'll change that and add a warning too.

lumag: done

tarcieri:
@lumag need to remove `expose-internals` here: https://github.com/RustCrypto/RSA/blob/master/.github/workflows/ci.yml#L70

lumag:
done

tarcieri (approved):

PR #307, opened 2023-04-25
Improved public key checks

### LINKED ISSUE #99: Checks performed in check_public seem insufficient

tarcieri:
Adds the following checks:

- `n` is odd
- `e` is odd
- `e` < `n`

Closes #99

### DIFF
--- a/src/key.rs
+++ b/src/key.rs
@@ -3,6 +3,7 @@ use core::ops::Deref;
 use num_bigint::traits::ModInverse;
 use num_bigint::Sign::Plus;
 use num_bigint::{BigInt, BigUint};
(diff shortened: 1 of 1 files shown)


dignifiedquire (commented):
thanks 👍

PR #310, opened 2023-04-26
Make `PaddingScheme`/`SignatureScheme` object safe

### LINKED ISSUE #261: Should `PaddingScheme` and `SignatureScheme` be object safe?

tarcieri:
This commit removes the `Rng` generic parameter, replacing it with `dyn CryptoRngCore`.

By eliminating the generic parameter from the signature, the traits become object safe.

This closes #261, which suggested these traits should be object safe. The types that impl these traits were previously a `PaddingScheme` enum which made it easy to select variants at runtime, but that functionality was lost when they were split apart with unifying traits.

Since these traits and the types that impl them effectively represent a "dynamic API" which allows potentially many key usages, making these traits object safe further enables that dynamism.

### DIFF
--- a/src/key.rs
+++ b/src/key.rs
@@ -14,10 +14,8 @@ use serde::{Deserialize, Serialize};
 use zeroize::Zeroize;
 
 use crate::algorithms::generate::generate_multi_prime_key_with_exp;
-use crate::dummy_rng::DummyRng;
 use crate::errors::{Error, Result};
 use crate::keytraits::{CRTValue, PrivateKeyParts, PublicKeyParts};
-
 use crate::padding::{PaddingScheme, SignatureScheme};
 
 /// Represents the public part of an RSA key.
@@ -364,7 +362,7 @@ impl RsaPrivateKey {
 
     /// Decrypt the given message.
     pub fn decrypt<P: PaddingScheme>(&self, padding: P, ciphertext: &[u8]) -> Result<Vec<u8>> {
-        padding.decrypt(Option::<&mut DummyRng>::None, self, ciphertext)
+        padding.decrypt(None, self, ciphertext)
     }
 
     /// Decrypt the given message.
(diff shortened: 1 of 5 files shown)


src/key.rs
```diff
@@ -364,7 +362,7 @@ impl RsaPrivateKey {
 
     /// Decrypt the given message.
     pub fn decrypt<P: PaddingScheme>(&self, padding: P, ciphertext: &[u8]) -> Result<Vec<u8>> {
-        padding.decrypt(Option::<&mut DummyRng>::None, self, ciphertext)
+        padding.decrypt(None, self, ciphertext)
```
tarcieri, line 365: A fringe benefit here is getting rid of some `DummyRng` usages. With a little more work it might be possible to completely get rid of it.

tarcieri:
FWIW, I went ahead and pushed up a branch based on this one which completely eliminates `DummyRng`: https://github.com/RustCrypto/RSA/commit/0c038fbf47a1a4e7f0074ad58bcd567ef612a005

tarcieri:
Hmm, there's one problem with this: you can't actually pass a trait object to e.g. `RsaPrivateKey::decrypt` because it accepts the `S` parameter by value instead of reference, so it can't be `?Sized`

tarcieri:
Seems this isn't actually possible/useful because of the need to move out of `self`. Closing.

PR #312, opened 2023-04-26
Refactor padding modes into submodules

tarcieri:
The padding mode modules have gotten quite large.

This commit refactors types into respective submodules, with the toplevel module defining the same-named padding schemes.

cc @lumag


tarcieri:
Note: this PR is just moving code around and should have only additive changes to re-exports, rather than any other API changes.

One thing it didn't refactor was the tests, although many of them are interdependent in a way it's hard to test anywhere but the toplevel.

PR #317, opened 2023-04-27
Replace `Deref` impl on `RsaPrivateKey` with `AsRef`

tarcieri:
The `RsaPrivateKey` type previously had a `Deref` impl providing access to the associated `RsaPublicKey`.

`Deref` is intended for "smart pointer types", i.e. container types which manage a (typically generic) inner type in some way. This doesn't seem like one of those cases.

`AsRef`, on the other hand, is for cheap reference conversions, which is exactly what's happening here, so it's a better fit and provides the same functionality (albeit explicitly rather than via deref coercion).

### DIFF
--- a/src/key.rs
+++ b/src/key.rs
@@ -1,8 +1,5 @@
 use alloc::vec::Vec;
-use core::{
-    hash::{Hash, Hasher},
-    ops::Deref,
-};
+use core::hash::{Hash, Hasher};
 use num_bigint::traits::ModInverse;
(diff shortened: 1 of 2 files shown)


tarcieri:
FYI: this is the last change I have in mind prior to a v0.9.0 release

cc @lumag

dignifiedquire (approved):

lumag:
@tarcieri please excuse me for missing several last PRs. I am at the conference now. They look good for me.

PR #325, opened 2023-05-03
Left pad signatures when encoding

tarcieri:
The `SignatureEncoding` impl added in `rsa` v0.9 (or more specifically, the `From<Signature>` impl for `Box<[u8]>` failed to properly left pad the signatures so they matched the modulus size.

This adds the appropriate padding to the signature encoder.

### DIFF
--- a/src/pkcs1v15/signature.rs
+++ b/src/pkcs1v15/signature.rs
@@ -4,6 +4,7 @@ pub use ::signature::{
     SignatureEncoding, Signer, Verifier,
 };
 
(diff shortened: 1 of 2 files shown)


baloo:
Thanks!

tarcieri:
I confirmed this fixes `RSASSA-PKCS1-v1_5` signatures when rebasing #326.

`RSASSA-PSS` signatures are still failing proptests, and I'm not sure why.

Anyway, going to go ahead and merge this.

PR #326, opened 2023-05-04
Initial proptests

tarcieri:
Adds proptests for RSASSA-PKCS1-v1_5 and RSA-PSS.

NOTE: these are currently expected to fail, and to be resolved by #325

### DIFF
--- a/tests/proptests.rs
+++ b/tests/proptests.rs
@@ -1,0 +1,42 @@
+//! Property-based tests.
+
+use proptest::prelude::*;
+use rand_chacha::ChaCha8Rng;
+use rand_core::SeedableRng;
+use rsa::{
+    pkcs1v15,
+    signature::{Keypair, SignatureEncoding, Signer, Verifier},
+    RsaPrivateKey,
+};
+use sha2::Sha256;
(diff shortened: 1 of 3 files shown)


tests/proptests.rs
```diff
@@ -0,0 +1,42 @@
...
+    // TODO(tarcieri): debug why these are failing
+    // #[test]
+    // fn pss_sign_roundtrip(private_key in private_key(), msg in any::<Vec<u8>>()) {
+    //     let signing_key = pss::SigningKey::<Sha256>::new(private_key);
+    //     let signature_bytes = signing_key.sign(&msg).to_bytes();
+    //
+    //     let verifying_key = signing_key.verifying_key();
+    //     let signature = pss::Signature::try_from(&*signature_bytes).unwrap();
+    //     prop_assert!(verifying_key.verify(&msg, &signature).is_ok());
+    // }
```
tarcieri, line 41: These are at least blowing up inside of `sign`, so it's not producing a signature that fails to verify at least.

Somewhat concerning though. Hopefully related to the small key size?

PR #340, opened 2023-06-17
fix doc example so that it works

spencerbart:


spencerbart:
I don't know if we should put a message saying that the sha2 feature needs to be enabled to get this example to work.

src/lib.rs
```diff
@@ -19,7 +19,7 @@
 //!
 //! ## OAEP encryption
 //! ```
-//! use rsa::{RsaPrivateKey, RsaPublicKey, Oaep};
+//! use rsa::{RsaPrivateKey, RsaPublicKey, Oaep, sha2};
```
tarcieri, line 22:
```suggestion
//! use rsa::{RsaPrivateKey, RsaPublicKey, Oaep, sha2::Sha256};
```

tarcieri:
> message saying that the sha2 feature needs to be enabled

sure

tarcieri:
Closing in favor of #381

Issue #343, opened 2023-06-27
Missing check in prime generation?


darconeous:
I was reading the [RSA key generation code](https://github.com/RustCrypto/RSA/blob/823bb231625c7713f856b9542ec91e1c1935e3d3/src/algorithms/generate.rs#L61..L112) and I couldn't seem to find the check where we make sure that $e$ is not a factor of $(q-1)$ or $(q-1)$.

Am I missing something? Without that check, it would seem that some small portion (~ $2/e$ ) of the generated keys would be invalid (the private key would be [incapable of properly decrypting or signing](https://eprint.iacr.org/2020/1059.pdf)).

I hope I'm just missing it.

tarcieri:
I believe it's here: https://github.com/RustCrypto/RSA/blob/823bb23/src/algorithms/generate.rs#L107

tarcieri:
See also: https://github.com/dignifiedquire/num-bigint/blob/2aade9d/src/algorithms/mod_inverse.rs#L12-L16

tarcieri:
Going to go ahead and close this. Reopen if you disagree.

tarcieri:
@dignifiedquire we might consider adding a comment about this to assist people looking for it in the future

PR #345, opened 2023-06-29
Add clarifying comment for totient check in key generation

tarcieri:
See #343 where it wasn't entirely clear


dignifiedquire (approved):
thanks

Issue #351, opened 2023-07-21
Undocumented breaking change in v0.9: `expose-internals` removed with no replacement


tchebb:
In v0.8 and prior, the `expose-internals` feature, when enabled, would export several low-level functions that generally shouldn't be used directly. Keeping those functions behind an "I know what I'm doing" feature flag was IMO a good compromise, since it gave the crate secure defaults but also let users who really need those functions access them without patching the crate.

Unfortunately, with v0.9 and #304, those functions were moved to various other places and hidden behind multiple layers of `pub(crate)`, meaning downstream users of rsa have no way at all to use them anymore (even though [this commit message](https://github.com/RustCrypto/RSA/pull/304/commits/d5e42fe33e2b4603f7e392f659d1ce8ca47aba61) acknowledges that "it is expected that the functions inside algorithms crates might be useful (and used) by other parties"). The `expose-internals` feature flag was removed too, but none of this was documented in `CHANGELOG.md`.

I now find myself unable to upgrade one of my projects from v0.8 to v0.9 because it relies on the raw `encrypt()` and `decrypt()` functions. The project (which unfortunately isn't open source, or I'd link to it) decrypts firmware images for a commercial piece of hardware, the format of which I reverse engineered. As such, I have no control over what crypto operations it uses: the encryption scheme was defined by the vendor, and my code needs to implement it as is—bad choices and all. One of the bad choices the vendor made was to encrypt blocks of data with no padding whatsoever, which is why I need `encrypt()`.

Would you accept a PR to bring `expose-internals` back? The functions themselves don't have to move back—`internals` could just re-export them from their new locations, and we could remove `pub(crate)` selectively on only the ones that make up the internal API. I personally only need `encrypt()` and `decrypt()`, so I'd be fine leaving the rest hidden unless/until someone demonstrates a compelling need for them.

PR #352, opened 2023-07-21
Document `expose-internals` removal, bring part of `internals` back as `hazmat`

tchebb:
As per the discussion in #351, some of these functions are useful to external users, but we want to make it absolutely clear that they're unsafe. Keep them behind a feature flag for that reason, and additionally add scary documentation inspired by what's [already in the aes crate](https://github.com/RustCrypto/block-ciphers/blob/master/aes/src/hazmat.rs).

Please don't squash when merging if possible; the commits are split appropriately as is.

This contribution is on behalf of my company.


tarcieri (approved):
Looks good, thanks

Issue #323, opened 2023-05-03
RSA-KEM support


tarcieri:
RSA-KEM is a key encapsulation mechanism for transmitting another cryptographic key (e.g. a symmetric key for a hybrid cryptosystem) between parties.

We have a trait-based abstraction over KEMs which provides a common API: https://github.com/RustCrypto/traits/tree/master/kem

It would be good to add (possibly optional/feature-gated) support for using the `kem` interface with RSA-KEM.

cc @rozbb

PR #355, opened 2023-08-09
add rsa_ext

Caisin:
pri key encrypt long string 
pub key decrypt long data

### DIFF
--- a/src/algorithms/pad.rs
+++ b/src/algorithms/pad.rs
@@ -8,7 +8,7 @@ use crate::errors::{Error, Result};
 
 /// Returns a new vector of the given length, with 0s left padded.
 #[inline]
-fn left_pad(input: &[u8], padded_len: usize) -> Result<Vec<u8>> {
+pub(crate) fn left_pad(input: &[u8], padded_len: usize) -> Result<Vec<u8>> {
     if input.len() > padded_len {
         return Err(Error::InvalidPadLen);
     }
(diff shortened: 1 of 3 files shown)


tarcieri:
I've mostly seen this referred to as RSA in Electronic CodeBook (ECB) mode.

It's a suboptimal way to use RSA as it needs to perform a modpow operation per "block", and generally discouraged.

A better option is to use [hybrid encryption](https://en.wikipedia.org/wiki/Hybrid_cryptosystem), which uses something like RSA-KEM to establish a shared symmetric key, then symmetric cryptography to actually encrypt the message.

[HPKE](https://github.com/rozbb/rust-hpke) provides a standard protocol for this, although unfortunately we don't yet provide RSA-KEM support: #323

Caisin:
I know what you mean
because this encryption and decryption method is provided in other programming languages
when we interact with other third-party systems, they use this way to verify encryption and decryption of apis,
in order to adapt to their api, we have to use this way to encrypt and decrypt,
I searched for a long time in crates.io and  github, but did not find,
so I implement it myself.
I believe that there will be others who have the same needs as me
so I create this pull request

tarcieri:
> because this encryption and decryption method is provided in other programming languages

That's not a justification for including this feature. The term for that is [cargo culting](https://en.wiktionary.org/wiki/cargo_culting).

I would personally prefer not to include such a feature in the `rsa` crate because:

- It is a non-standard construction without an associated standards document like an RFC
- It is badly designed and including it is something of an endorsement
- Hybrid encryption should be the preferred solution to this problem

However, feel free to publish your own crate that implements this construction.

PR #357, opened 2023-08-18
Remove unused dependencies

psychon:
I came across these unused dependencies and wanted to remove them "just because".


tarcieri (approved):

tarcieri:
Thanks! Perhaps we should lint for this

PR #359, opened 2023-08-19
Pin serde to <1.0.172

newpavlov:
`serde v1.0.172` and later include pre-compiled binaries which is a security hazard. So until the decision gets reverted, I believe it's worth to pin upper version of `serde`. This approach may cause issues if a different crate in someone's dependency tree will depend on a post-1.0.172 version of `serde`, but I think this issue is small enough when compared to the security concerns. Also, a number of other crates in the ecosystem follow this approach, so we are not alone.

More information and discussion about the `serde` change can be found in https://github.com/serde-rs/serde/issues/2538.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,13 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## UNRELEASED
+### Changed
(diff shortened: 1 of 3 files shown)


dignifiedquire (approved):
sad, but makes sense

tarcieri:
Isn't the offending binary in the `serde_derive` crate?

newpavlov:
`serde` uses pinned version of `serde_derive`, i.e. `serde v1.0.172` uses strictly only `serde_derive v1.0.172`

newpavlov:
@dignifiedquire 
I will merge this PR, but will not do a release and will leave it up to your discretion.

PR #378, opened 2023-10-26
CI: pin nightly; test minimal-versions on stable

tarcieri:
- Pins to nightly-2023-10-01 to prevent regressions
- On `minimal-versions` only uses nightly to resolve versions, and runs tests on stable instead

This is a workaround for build failures we're seeing due to a recent nightly regression. See:

https://github.com/RustCrypto/RSA/actions/runs/6655788342/job/18086838437?pr=377

https://github.com/RustCrypto/RSA/actions/runs/6655788342/job/18086838800?pr=377

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -63,10 +63,9 @@ jobs:
     steps:
       - uses: actions/checkout@v4
       - uses: RustCrypto/actions/cargo-cache@master
-      - uses: dtolnay/rust-toolchain@master
-        with:
-          toolchain: nightly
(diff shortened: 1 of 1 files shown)


newpavlov (approved):
It looks good as a temporary measure. But I think we should revert it after the regression is fixed. Do we know if someone create a Rust issue for it?

UPD: The issue is https://github.com/rust-lang/rust/issues/117120 and it looks like it was already fixed. We probably can simply wait for a future Nightly release with the fix.

tarcieri:
I think it would be good to change `minimal-versions` to always test using `stable` after using `nightly` to resolve the lockfile

Issue #19, opened 2019-04-30
modpow implementation is not constant-time


phayes:
Hi there, 

I'm the author of sidefuzz (https://github.com/phayes/sidefuzz) and I have found what appears to be variable-time behavior in the `rsa::internals::encrypt()` function.   Specifically, `rsa::internals::encrypt()` appears to be variable-time in relation to the message. Note that I haven't worked this up into an actual exploit, but merely demonstrated that this function isn't constant-time in relation to the message inputed. 

Specifically, the message `20d90c8af42aac9b1ee53dc9a0187201` takes 549894 instructions to encrypt, while the message `5a28cec68d47f6fe3b1df54c9f320f6d` takes 552427 instruction to encrypt.  This is a difference of 2533 instructions, or about 0.5%.   So it's a slight difference, but probably exploitable with sufficient sampling. 

I have crated a fuzzing targets here: https://github.com/phayes/sidefuzz-targets

You can confirm this difference with the [sidefuzz](https://github.com/phayes/sidefuzz) tool like so:

```
sidefuzz check ./target/wasm32-unknown-unknown/release/rsa_encrypt_message.wasm 5a28cec68d47f6fe3b1df54c9f320f6d 20d90c8af42aac9b1ee53dc9a0187201
samples: 20000, t-value: 219771.0790572351, confidence: 100%
Found timing difference of 2533 instructions between these two inputs with 100% confidence:
input 1: 5a28cec68d47f6fe3b1df54c9f320f6d (552427 instructions)
input 2: 20d90c8af42aac9b1ee53dc9a0187201 (549894 instructions)
```

My first suspicion was that this was due to `num_bigint_dig::BigUint::from_bytes_be()` being variable-time, but fuzzing that function specifically results in what appears to be constant-time behavior.  So I'm not actually sure where the problem is.

newpavlov:
IIRC RSA blinding (which I believe is used in this implementation) uses random number, so while execution time is still variable, it does not correlate with a protected secret.

phayes:
Hi @newpavlov , 

I don't think that's the case here. The [fuzzing target](https://github.com/phayes/sidefuzz-targets/blob/master/rsa_encrypt_message/src/lib.rs) is directly testing this function:

```
/// Raw RSA encryption of m with the public key. No padding is performed.
#[inline]
pub fn encrypt<K: PublicKey>(key: &K, m: &BigUint) -> BigUint {
    m.modpow(key.e(), key.n())
```

[...]

PR #391, opened 2023-11-28
README.md: add info about Marvin Attack (RUSTSEC-2023-0071)

tarcieri:
References:
- https://github.com/RustCrypto/RSA/issues/19
- https://rustsec.org/advisories/RUSTSEC-2023-0071.html
- https://people.redhat.com/~hkario/marvin/

### DIFF
--- a/README.md
+++ b/README.md
@@ -65,16 +65,19 @@ There will be three phases before `1.0` 🚢 can be released.
     - [ ] Fuzz testing
     - [ ] Security Audits
 
-## Security Notes
+## ⚠️Security Warning
 
 This crate has received one [security audit by Include Security][audit], with
 only one minor finding which has since been addressed.
 
 See the [open security issues] on our issue tracker for other known problems.
 
-Notably the implementation of [modular exponentiation is not constant time],
+~~Notably the implementation of [modular exponentiation is not constant time],
 but timing variability is masked using [random blinding], a commonly used
-technique.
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -65,16 +65,19 @@ There will be three phases before `1.0` 🚢 can be released.
...
 only one minor finding which has since been addressed.
 
 See the [open security issues] on our issue tracker for other known problems.
 
-Notably the implementation of [modular exponentiation is not constant time],
+~~Notably the implementation of [modular exponentiation is not constant time],
 but timing variability is masked using [random blinding], a commonly used
-technique.
+technique.~~ This crate is vulnerable to the [Marvin Attack] which could enable
+private key recovery by a network attacker (see [RUSTSEC-2023-0071]).
```
tarcieri, line 78: I should probably note that while the "Marvin Attack" itself only permits signing or decryption and not actual recovery of the public key, the timing variability of our core `rsa_decrypt` function seems severe enough that it's something worth warning about (also being able to do things the private key enables without recovering the private key isn't that much better of a situation)

tomato42: Correct. While Marvin Attack isn't about extraction of the RSA private key, lack of blinding on the RSA decryption operation and non-constant time numerical library used for the modular exponentiation mean that private key extraction is very likely possible. See references from Section 2 of https://datatracker.ietf.org/doc/draft-kario-rsa-guidance/

PR #393, opened 2023-11-28
pss: expose a `get_default_pss_signature_algo_id` helper

baloo:
This returns the `AlgorithmIdentifierOwned` assuming the default situation where the salt_len is the `output_size` of the `Digest`.

This is meant to be used by HSM where the private key is hidden behind an abstraction but to still expose a `DynSignatureAlgorithmIdentifier` trait.

### DIFF
--- a/src/pss.rs
+++ b/src/pss.rs
@@ -227,6 +227,15 @@ fn sign_pss_with_salt_digest<T: CryptoRngCore + ?Sized, D: Digest + FixedOutputR
     )
 }
 
+/// Returns the [`AlgorithmIdentifierOwned`] associated with PSS signature using a given digest.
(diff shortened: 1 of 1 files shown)


baloo:
Usage:
 - yubihsm: https://github.com/iqlusioninc/yubihsm.rs/blob/2a7c2ca627db4246d658d3f78ceaaab7509f1df4/src/rsa/pss/signer.rs#L94
 - pkcs#11: https://github.com/parallaxsecond/rust-cryptoki/blob/819a72bf9e0e78a3a7dc0c6a0ac25530629f1dd5/cryptoki-rustcrypto/src/rsa/pss.rs#L109

tarcieri (approved):

Issue #51, opened 2020-06-11
core-only support


roblabla:
Right now, there is a PR open (#22) that will allow the RSA crate to be used in core+alloc environments (e.g. environments without libstd, but where a dynamic allocator is available).

The next step is for the RSA crate to support core-only environments. This would require multiple changes: `num-bigint-dig` would need to be represented either as a stack-allocated array (GenericArray? ArrayVec?) or a slice reference (wooo lifetimes) instead of a heap-allocated vector. Furthermore, we'll need to make extra sure that all the operations are bounded within a certain numeric range.

dignifiedquire:
it will be hard, but very valuable to be able to make the bigint lib usable with genericarray,

tarcieri:
You could potentially use [`heapless::Vec`](https://docs.rs/heapless/0.5.5/heapless/struct.Vec.html) to simplify usage of `GenericArray`.

Unfortunately they haven't updated to v0.14 yet, which is annoying:

https://github.com/japaric/heapless/issues/166

dignifiedquire:
Yes `heapless::Vec` has some nice utilities that we would need. I would love to keep the ability for returning larger numbers, eg multiply that returnst the full result and not just the truncated, might make sense to implement the std operations like std lib, with overflow panic, and having `wrapping`, `checked` and a new variant sth like `full` which would return the fully expanded version

newpavlov:
>I would love to keep the ability for returning larger numbers, eg multiply that returnst the full result and not just the truncated

IIUC it should be possible to do completely on stack, with something like:
```rust
fn mul<const N: usize>(a: [Word; N], b: [Word; N]) -> [Word; 2*N] { .. }
```

tarcieri:
For what it's worth, I started working on a crate to abstract over the details between containers like `Vec` and `heapless::Vec` (or `arrayvec` if so desired), with all operations fallible for use with fixed-sized containers:

- https://github.com/RustCrypto/utils/tree/master/collectable
- https://docs.rs/collectable/0.0.1/collectable/

Not using it for anything yet and it could probably use some work, but I think it had a similar inspiration to this sort of use case.

dignifiedquire:
@newpavlov it is possible, but as we don't have const generics yet this gets quite ugly and complex quickly, especially if your interfaces have to rely on the certain size, as they would for RSA given a fixed bitwitdth

tarcieri:
[...]

PR #394, opened 2023-11-29
feat: switch from num-bigint-dig to crypto-bigint 

dignifiedquire:
~~Very, very WIP~~

Not anymore, this is ready for review. 
Replaces all usage of `num-bigint-dig` based `BigInt` usage with the new `crypto-bigint` crate, using `BoxedUint`

Current known issue is that we do have a performance regression, which will be able to get rid of over time:

```
# crypto-bigint

# macbook m1
test bench_rsa_2048_pkcsv1_decrypt      ... bench:   7,184,387.50 ns/iter (+/- 425,598.69)
test bench_rsa_2048_pkcsv1_sign_blinded ... bench:  13,453,579.10 ns/iter (+/- 686,276.31)

# AMD
test bench_rsa_2048_pkcsv1_decrypt      ... bench:   9,260,832.80 ns/iter (+/- 30,013.38)
test bench_rsa_2048_pkcsv1_sign_blinded ... bench:  16,610,079.40 ns/iter (+/- 251,292.53)

# master

# macbook m1
test bench_rsa_2048_pkcsv1_decrypt      ... bench:   1,117,479.15 ns/iter (+/- 31,334.30)
test bench_rsa_2048_pkcsv1_sign_blinded ... bench:   1,337,437.55 ns/iter (+/- 88,624.39)

# AMD
test bench_rsa_2048_pkcsv1_decrypt      ... bench:   1,414,348.80 ns/iter (+/- 12,585.71)
test bench_rsa_2048_pkcsv1_sign_blinded ... bench:   1,685,650.00 ns/iter (+/- 11,105.71)
```

## TODOs

- [x] switch internal storage for `RsaPrivateKey`
- [x] switch internal storage for `RsaPublicKey`
- [x] switch all code to use the new `decrypt` implementation
- [x] update public traits using `BigUint` to return owned versions
- [x] fix blinding implementation
- [x] switch decryption algorithm with precompute to use crypto-bigint ops
- [x] go through other algorithms and update what can be done without having primality checks implemented
- [x] review & update code for constant time operation
- [x] review & update code for performance
- [x] benchmarks


src/algorithms/rsa.rs
```diff
@@ -151,38 +174,40 @@ pub fn rsa_decrypt_and_check<R: CryptoRngCore + ?Sized>(
...
         }
-        ir = r.clone().mod_inverse(key.n());
-        if let Some(ir) = ir {
-            if let Some(ub) = ir.into_biguint() {
-                unblinder = ub;
-                break;
-            }
+        ir = r.inv_mod(key.n());
+
+        // TODO: constant time?
```
tarcieri, line 200: Ordinarily I'd suggest `CtOption::map` here but it doesn't work on `BoxedUint` due to a `Copy` bound required for `ConditionallySelectable`: https://github.com/dalek-cryptography/subtle/issues/94#issuecomment-1829905653

We might need to add a little library of helper functions to work around it, however it's complicated by the need to respect the precision of the original value. I could maybe PR something to `crypto-bigint` as we might end up needing it there.

tarcieri: I added some constant time combinators for `BoxedUint`, but I now realize that they wouldn't help in this case:

https://github.com/RustCrypto/crypto-bigint/pull/352

However, they're now available if you do end up needing them.

src/algorithms/rsa.rs
```diff
@@ -36,14 +37,12 @@ pub fn rsa_encrypt<K: PublicKeyParts>(key: &K, m: &BigUint) -> Result<BigUint> {
 pub fn rsa_decrypt<R: CryptoRngCore + ?Sized>(
     mut rng: Option<&mut R>,
     priv_key: &impl PrivateKeyPartsNew,
-    c_orig: &BigUint,
+    c: &BoxedUint,
```
tarcieri, line 40: It might be worth leaving this as-is to preserve backwards compatibility, so we could potentially ship a v0.9.x patch release focused just on making `rsa_decrypt` constant-time (in theory)

dignifiedquire: the public traits already have a breaking change, so I figured we might as well chnage the hazmat apis as well

tarcieri:
@dignifiedquire FYI: https://docs.rs/crypto-primes/0.6.0-pre.0/crypto_primes/

bradjc:
Is there any plan to support the `Uint` non-alloc type?

tarcieri:
@bradjc it should be possible in the future (see #51) and `crypto-primes` has been updated to support both with generic algorithms.

However, it will be difficult, so a first pass should probably focus on getting a heap-allocated implementation working.

nicolasauler:
Is there work still happening here or is this issue blocked?
How can the community help?
My pipelines are still breaking in the security audit due to the Marvin Attack vuln

tarcieri:
I can potentially pick up this PR or https://github.com/RustCrypto/crypto-bigint/pull/403

> How can the community help?

@nicolasauler I haven't had time to check out #400 yet, but if you'd like to confirm it's working as expected that would be helpful. It implements the tooling for the attack and so far we've managed to confirm that while this PR is an improvement it doesn't seem to entirely close the sidechannel.

> My pipelines are still breaking in the security audit due to the Marvin Attack vuln

You should be able to ignore `RUSTSEC-2023-0071` in e.g. `.cargo/audit.toml` for now (provided you're using `cargo-audit`).

But yes, we're in the same boat where several of the @RustCrypto repos are currently labeled "insecure" due to this vulnerability.

bradjc:
Is it reasonable to say the implementation is fairly close, and that the remaining items would only result in minor fixes?

tarcieri:
I can potentially pick up this branch if @dignifiedquire is too busy. It looks like there's still a decent number of things to do, though the tests appear to be broken due to a `[patch.crates-io]` directive pointing at the local filesystem

dignifiedquire:
> though the tests appear to be broken due to a `[patch.crates-io]` directive pointing at the local filesystem

this is just pointing at this branch on my machine: https://github.com/RustCrypto/crypto-bigint/tree/feat-expand-mul

dignifiedquire:
The big changes needed to make this performant enough, is the work I started to allow different width of bigints to be operated on, as we are currently wasting a lot of time in operating on much larger values than we need to. Everything else is mostly done.

tarcieri:
@dignifiedquire perhaps we should get an initial working implementation landed then follow up on performance separately?

dignifiedquire:
> @dignifiedquire perhaps we should get an initial working implementation landed then follow up on performance separately?

We could, but performance of this crate is already not great, and it goes down by at least a factor of 2 :(

tarcieri:
Better than a sidechannel vulnerability that leaks the private key!

MasterAwesome:
> @bradjc it should be possible in the future (see #51) and `crypto-primes` has been updated to support both with generic algorithms.
> 
> However, it will be difficult, so a first pass should probably focus on getting a heap-allocated implementation working.

Is there work required in any of the dependencies that are blocking non-alloc types from being used? If there is, I can try a stab at it parallelly.

tarcieri:
@MasterAwesome the dependencies are largely ready to go, including `crypto-primes`.

The hard part will be making the implementation of `rsa` generic around the underlying integer type, and that's something we can't really even begin to explore until we have it fully migrated to `crypto-bigint` in any capacity.

tarcieri:
@MasterAwesome in terms of something that would generally help with the migration, there are various avenues for `crypto-bigint` performance improvements to be explored: https://github.com/RustCrypto/crypto-bigint/pull/403

dignifiedquire:
Update:

## done

- merged latest `master` p
- switched back to fixed sizes 
- updated to crpyto-bigint alpha.12

# next steps

- figure out why 3 tests are failing
- integrate prime number generation

dignifiedquire:
Almost all tests are broken, but `num-bigint` is not used anymore, all code is moved to crypto-bigint now

src/encoding.rs
```diff
@@ -34,12 +35,13 @@ impl TryFrom<pkcs8::PrivateKeyInfo<'_>> for RsaPrivateKey {
         if pkcs1_key.version() != pkcs1::Version::TwoPrime {
             return Err(pkcs1::Error::Version.into());
         }
-
-        let n = BigUint::from_bytes_be(pkcs1_key.modulus.as_bytes());
-        let e = BigUint::from_bytes_be(pkcs1_key.public_exponent.as_bytes());
-        let d = BigUint::from_bytes_be(pkcs1_key.private_exponent.as_bytes());
-        let prime1 = BigUint::from_bytes_be(pkcs1_key.prime1.as_bytes());
-        let prime2 = BigUint::from_bytes_be(pkcs1_key.prime2.as_bytes());
+        let bits = 512; // TODO: read from data
```
tarcieri, line 38: You should be able to use `pkcs1_key.modulus.as_bytes().len() * 8` for this, I think?

pinkforest:
It seems some trait impls are missing / some bound shenanigans

Fixing for sede De/Serialize for BoxedUint - some TODO left there
- https://github.com/RustCrypto/crypto-bigint/pull/587

Rest seem bounds issues, I'll send a PR

src/algorithms/generate.rs
```diff
@@ -78,8 +76,9 @@ pub(crate) fn generate_multi_prime_key_with_exp<R: CryptoRngCore + ?Sized>(
         }
 
         for (i, prime) in primes.iter_mut().enumerate() {
-            *prime = rng.gen_prime(todo / (nprimes - i));
-            todo -= prime.bits();
+            let bits = (todo / (nprimes - i)) as u32;
+            *prime = generate_prime_with_rng(rng, bits, bits);
```
pinkforest, line 80: The bit precision (2nd arg) here seems to be wrong ?

The tests fails because there is this behind:
`let mut random = T::random_bits_with_precision(rng, bit_length, bits_precision);`

random.bits_precision() = 128, bits_precision = 85, bit_length = 85, bits_precision = 85
random.bits_precision() = 512, bits_precision = 512, bit_length = 512, bits_precision = 512
random.bits_precision() = 64, bits_precision = 6, bit_length = 6, bits_precision = 6

85 should be 128 - 6 should be 64 hmmm

pinkforest, line 80:
```suggestion
            let bits = (todo / (nprimes - i)) as u32;
            let bits_precision = BoxedUint::zero_with_precision(bits).bits_precision();
            *prime = generate_prime_with_rng(rng, bits, bits);
```

One way to get it :woman_shrugging:

src/oaep.rs
```diff
@@ -331,13 +344,14 @@ mod tests {
...
-                BigUint::parse_bytes(b"00da4817ce0089dd36f2ade6a3ff410c73ec34bf1b4f6bda38431bfede11cef1f7f6efa70e5f8063a3b1f6e17296ffb15feefa0912a0325b8d1fd65a559e717b5b961ec345072e0ec5203d03441d29af4d64054a04507410cf1da78e7b6119d909ec66e6ad625bf995b279a4b3c5be7d895cd7c5b9c4c497fde730916fcdb4e41b", 16).unwrap()
+                BoxedUint::from_be_hex("00f827bbf3a41877c7cc59aebf42ed4b29c32defcb8ed96863d5b090a05a8930dd624a21c9dcf9838568fdfa0df65b8462a5f2ac913d6c56f975532bd8e78fb07bd405ca99a484bcf59f019bbddcb3933f2bce706300b4f7b110120c5df9018159067c35da3061a56c8635a52b54273b31271b4311f0795df6021e6355e1a42e61", bits).unwrap(),
+                BoxedUint::from_be_hex("00da4817ce0089dd36f2ade6a3ff410c73ec34bf1b4f6bda38431bfede11cef1f7f6efa70e5f8063a3b1f6e17296ffb15feefa0912a0325b8d1fd65a559e717b5b961ec345072e0ec5203d03441d29af4d64054a04507410cf1da78e7b6119d909ec66e6ad625bf995b279a4b3c5be7d895cd7c5b9c4c497fde730916fcdb4e41b", bits).unwrap()
```
pinkforest, line 354:
```suggestion                BoxedUint::from_be_hex("00f827bbf3a41877c7cc59aebf42ed4b29c32defcb8ed96863d5b090a05a8930dd624a21c9dcf9838568fdfa0df65b8462a5f2ac913d6c56f975532bd8e78fb07bd405ca99a484bcf59f019bbddcb3933f2bce706300b4f7b110120c5df9018159067c35da3061a56c8635a52b54273b31271b4311f0795df6021e6355e1a42e61", 128).unwrap(),
                BoxedUint::from_be_hex("00da4817ce0089dd36f2ade6a3ff410c73ec34bf1b4f6bda38431bfede11cef1f7f6efa70e5f8063a3b1f6e17296ffb15feefa0912a0325b8d1fd65a559e717b5b961ec345072e0ec5203d03441d29af4d64054a04507410cf1da78e7b6119d909ec66e6ad625bf995b279a4b3c5be7d895cd7c5b9c4c497fde730916fcdb4e41b", 128).unwrap()
```

src/algorithms/rsa.rs
```diff
@@ -305,19 +318,19 @@ pub(crate) fn compute_private_exponent_carmicheal(
...
+        let p = BoxedUint::from_be_hex("00f827bbf3a41877c7cc59aebf42ed4b29c32defcb8ed96863d5b090a05a8930dd624a21c9dcf9838568fdfa0df65b8462a5f2ac913d6c56f975532bd8e78fb07bd405ca99a484bcf59f019bbddcb3933f2bce706300b4f7b110120c5df9018159067c35da3061a56c8635a52b54273b31271b4311f0795df6021e6355e1a42e61", bits).unwrap();
+        let q = BoxedUint::from_be_hex("00da4817ce0089dd36f2ade6a3ff410c73ec34bf1b4f6bda38431bfede11cef1f7f6efa70e5f8063a3b1f6e17296ffb15feefa0912a0325b8d1fd65a559e717b5b961ec345072e0ec5203d03441d29af4d64054a04507410cf1da78e7b6119d909ec66e6ad625bf995b279a4b3c5be7d895cd7c5b9c4c497fde730916fcdb4e41b", bits).unwrap();
```
pinkforest, line 331:
```suggestion
        let p = BoxedUint::from_be_hex("00f827bbf3a41877c7cc59aebf42ed4b29c32defcb8ed96863d5b090a05a8930dd624a21c9dcf9838568fdfa0df65b8462a5f2ac913d6c56f975532bd8e78fb07bd405ca99a484bcf59f019bbddcb3933f2bce706300b4f7b110120c5df9018159067c35da3061a56c8635a52b54273b31271b4311f0795df6021e6355e1a42e61", 128).unwrap();
        let q = BoxedUint::from_be_hex("00da4817ce0089dd36f2ade6a3ff410c73ec34bf1b4f6bda38431bfede11cef1f7f6efa70e5f8063a3b1f6e17296ffb15feefa0912a0325b8d1fd65a559e717b5b961ec345072e0ec5203d03441d29af4d64054a04507410cf1da78e7b6119d909ec66e6ad625bf995b279a4b3c5be7d895cd7c5b9c4c497fde730916fcdb4e41b", 128).unwrap();
```

src/algorithms/rsa.rs
```diff
@@ -154,123 +140,149 @@ pub fn rsa_decrypt_and_check<R: CryptoRngCore + ?Sized>(
...
-pub fn recover_primes(n: &BigUint, e: &BigUint, d: &BigUint) -> Result<(BigUint, BigUint)> {
+pub fn recover_primes(
+    n: &NonZero<BoxedUint>,
+    e: u64,
+    d: &BoxedUint,
+) -> Result<(BoxedUint, BoxedUint)> {
     // Check precondition
-    let two = BigUint::from_u8(2).unwrap();
-    if e <= &two.pow(16u32) || e >= &two.pow(256u32) {
+    if e <= 2u64.pow(16) || e >= 2u64.pow(256) {
```
pinkforest, line 206: This overflows

src/algorithms/rsa.rs
```diff
@@ -154,123 +140,149 @@ pub fn rsa_decrypt_and_check<R: CryptoRngCore + ?Sized>(
...
     }
 
-    let mut totient = BigUint::one();
+    let mut totient = BoxedUint::one();
 
     for prime in primes {
-        totient *= prime - BigUint::one();
+        totient = totient * (prime - &BoxedUint::one());
     }
+    let totient = Odd::new(totient).unwrap();
```
pinkforest, line 280: panic - totient is Even

src/algorithms/rsa.rs
```diff
@@ -154,123 +140,149 @@ pub fn rsa_decrypt_and_check<R: CryptoRngCore + ?Sized>(
...
     }
 
     // 1. Let a = (de – 1) × GCD(n – 1, de – 1).
-    let one = BigUint::one();
-    let a = Zeroizing::new((d * e - &one) * (n - &one).gcd(&(d * e - &one)));
+    let one = BoxedUint::one();
+    let e = BoxedUint::from(e);
+
+    let a1 = d * &e - &one;
+    let a2 = (n.as_ref() - &one).gcd(&(d * e - &one)).unwrap();
```
pinkforest, line 215: BY-GCD left side Odd but Odd<modulus>-1 == Even in formula - [crypto-bigint/590](https://github.com/RustCrypto/crypto-bigint/issues/590)

src/algorithms/generate.rs
dignifiedquire: applied manually

dignifiedquire:
Most tests are passing at this point, outstanding failures seem to fall into two categories

- reconstructing keys fails, because we can't do gcd on even values yet
- the encoding/decoding is slightly different and so reading the test vectors or comparisons are failing there

tarcieri:
@dignifiedquire I think it'd be good to land a spike / PoC of this even if all the algorithms aren't yet implemented, since it would unblock splitting up the work on (re)implementing those

gogo2464:
@tarcieri can I be assigned to the poc please? I would like to gain experience.

gogo2464:
EDIT: I did not undertand the message yet. If the exploit is already made, yes I can read it.

gogo2464:
I can see difference but I would like a tool or a procedure to analyze the time spent. I am currently working with `callgrind` and `kcachegrind`.

dignifiedquire:
@tarcieri most things are working now 🎉 

I would appreciate some help with debugging the last failures, seems the proptests are discovering some roundtrip issues in the encoding/decoding and the last regular test that is a problem is dealign with a `2049bit` key

dignifiedquire:
and also feel free to start to review the code/fix up anything you think needs improving

kjvalencik:
This is awesome! I'm excited to see this PR progressing so much.

> seems the proptests are discovering some roundtrip issues in the encoding/decoding and the last regular test that is a problem is dealign with a 2049bit key

I looked into this for a few minutes and these two appear to be different cases of the same problem. `crypto-bigint` uses 64-bit limbs, which requires it to round higher on number of bytes. For example, with a 2049-bit key it's 264 instead of 257; leading to more padding in the serialized version. This may be the same thing causing round-trip problems.

src/algorithms/generate.rs
```diff
@@ -13,10 +11,10 @@ use crate::{
 };
 
 pub struct RsaPrivateKeyComponents {
-    pub n: BigUint,
-    pub e: BigUint,
-    pub d: BigUint,
-    pub primes: Vec<BigUint>,
+    pub n: Odd<BoxedUint>,
+    pub e: u64,
```
Fethbita, line 15: I don't think it's good to limit `e` to 64 bits. See [RFC 8017 Section 3.1](https://datatracker.ietf.org/doc/html/rfc8017#section-3.1):

> In a valid RSA public key, the RSA modulus n is a product of u distinct odd primes r_i, i = 1, 2, ..., u, where u >= 2, and the RSA public exponent e is an integer between 3 and n - 1 satisfying GCD(e,\lambda(n)) = 1, where \lambda(n) = LCM(r_1 - 1, ..., r_u - 1). By convention, the first two primes r_1 and r_2 may also be denoted p and q, respectively.

And [FIPS 186-4 page 52](https://nvlpubs.nist.gov/nistpubs/FIPS/NIST.FIPS.186-4.pdf#page=62)

> The exponent e **shall** be an odd positive integer such that:
>        2¹⁶ < e < 2²⁵⁶.

It should still be `BoxedUint`.

zeerooth:
I tried testing this PR in a `no_std` environment, but it doesn't compile, as RSA pulls `getrandom` via `crypto-bigint` even if I disable default features.

```
rsa = { git = "https://github.com/RustCrypto/RSA.git", branch = "const-crypto-biguint", default-features = false }
```

```
    rsa v0.10.0-pre.2 (https://github.com/RustCrypto/RSA.git?branch=const-crypto-biguint#74d31971)
    ├── const-oid v0.10.0-rc.0
    ├── crypto-bigint v0.6.0-rc.5
    │   ├── num-traits v0.2.19
    │   │   [build-dependencies]
    │   │   └── autocfg v1.4.0
    │   ├── rand_core v0.6.4
    │   │   └── getrandom v0.2.15
    │   │       └── cfg-if v1.0.0
    │   ├── subtle v2.6.1
    │   └── zeroize v1.8.1
```

```
error: target is not supported, for more information see: https://docs.rs/getrandom/#unsupported-targets
   --> /home/----/.cargo/registry/src/index.crates.io-6f17d22bba15001f/getrandom-0.2.15/src/lib.rs:347:9
    |
347 | /         compile_error!("target is not supported, for more information see: \
348 | |                         https://docs.rs/getrandom/#unsupported-targets");
    | |________________________________________________________________________^

```

src/algorithms/generate.rs
```diff
@@ -1,10 +1,8 @@
 //! Generate prime components for the RSA Private Key
 
 use alloc::vec::Vec;
-use num_bigint::{BigUint, RandPrime};
-#[allow(unused_imports)]
-use num_traits::Float;
```
baloo, line 6: [`Float::ln`](https://docs.rs/num-traits/latest/num_traits/float/trait.Float.html#tymethod.ln) was used to provide support for f64::ln on non-std (used line 45).

mepi262:
@dignifiedquire
Any update on this pull request?

dignifiedquire:
> @dignifiedquire Any update on this pull request?

no, haven't had time to fix the afformentioned issues yet

mepi262:
@dignifiedquire
Thank you for your response !
I hope your another project will be succeed and become be able to concentrate on this pull request.

src/algorithms/generate.rs
dignifiedquire: thanks for pointing this out, our default safe constants restricting it, but you are right, it should be possible for larger exponents. this is now fixed

dignifiedquire:
@tarcieri current benchmarks

```
# crypto-bigint

# macbook m1
test bench_rsa_2048_pkcsv1_decrypt      ... bench:   7,184,387.50 ns/iter (+/- 425,598.69)
test bench_rsa_2048_pkcsv1_sign_blinded ... bench:  13,453,579.10 ns/iter (+/- 686,276.31)

# AMD
test bench_rsa_2048_pkcsv1_decrypt      ... bench:   9,260,832.80 ns/iter (+/- 30,013.38)
test bench_rsa_2048_pkcsv1_sign_blinded ... bench:  16,610,079.40 ns/iter (+/- 251,292.53)

# master

# macbook m1
test bench_rsa_2048_pkcsv1_decrypt      ... bench:   1,117,479.15 ns/iter (+/- 31,334.30)
test bench_rsa_2048_pkcsv1_sign_blinded ... bench:   1,337,437.55 ns/iter (+/- 88,624.39)

# AMD
test bench_rsa_2048_pkcsv1_decrypt      ... bench:   1,414,348.80 ns/iter (+/- 12,585.71)
test bench_rsa_2048_pkcsv1_sign_blinded ... bench:   1,685,650.00 ns/iter (+/- 11,105.71)
```

dignifiedquire:
CI is finally green again, time for lots of review and cleanup

src/algorithms/generate.rs
dignifiedquire: thanks, fixed now

dignifiedquire:
@zeerooth `nostd` builds should be working again

src/algorithms/generate.rs
```diff
@@ -44,7 +42,10 @@ pub(crate) fn generate_multi_prime_key_with_exp<R: CryptoRngCore + ?Sized>(
         let prime_limit = (1u64 << (bit_size / nprimes) as u64) as f64;
 
         // pi aproximates the number of primes less than prime_limit
+        #[cfg(feature = "std")]
         let mut pi = prime_limit / (prime_limit.ln() - 1f64);
+        #[cfg(not(feature = "std"))]
+        let mut pi = prime_limit / (libm::logf(prime_limit as f32) as f64 - 1f64);
```
dignifiedquire, line 48: this is not entirely correct or optimal, but I am not sure what other way there is

tarcieri: @fjarri is there anything in `crypto-primes` for this?

fjarri: No, nothing like that currently

tarcieri:
@dignifiedquire seems like you can remove "Very, very WIP" now

dignifiedquire:
> @dignifiedquire seems like you can remove "Very, very WIP" now

🤣 yeah, finally

Snowiiii:
@dignifiedquire how it's going ?

tarcieri:
He's waiting on review, which I'll try to do soon

Cargo.toml
```diff
@@ -13,25 +13,26 @@ readme = "README.md"
...
+subtle = { version = "2.6.1", default-features = false }
 digest = { version = "=0.11.0-pre.9", default-features = false, features = ["alloc", "oid"] }
 pkcs1 = { version = "0.8.0-rc.0", default-features = false, features = ["alloc", "pkcs8"] }
 pkcs8 = { version = "0.11.0-rc.0", default-features = false, features = ["alloc"] }
 signature = { version = "=2.3.0-pre.4", default-features = false, features = ["alloc", "digest", "rand_core"] }
 spki = { version = "0.8.0-rc.1", default-features = false, features = ["alloc"] }
 zeroize = { version = "1.5", features = ["alloc"] }
+crypto-bigint = { version = "0.6.0-rc.6", default-features = false, features = ["zeroize", "alloc"] }
+crypto-primes = { version = "0.6.0-pre.2", default-features = false }
+libm = "0.2"
```
tarcieri, line 27: IMO it would be nice to avoid this, especially if its only used for `no_std` prime counting

README.md
```diff
@@ -81,7 +81,7 @@ You can follow our work on mitigating this issue in [#390].
 
 ## Minimum Supported Rust Version (MSRV)
 
-This crate supports Rust 1.72 or higher.
+This crate supports Rust 1.73 or higher.
```
tarcieri, line 84: This is `crypto-bigint`'s latest MSRV:

```suggestion
This crate supports Rust 1.83 or higher.
```

src/lib.rs
```diff
@@ -222,7 +223,6 @@ extern crate alloc;
 #[cfg(feature = "std")]
 extern crate std;
 
-pub use num_bigint::BigUint;
```
tarcieri, line 225: Perhaps consider a `BoxedUint` re-export here, possibly using a shorter name:

```rust
pub use crypto_bigint::BoxedUint as Uint;
```

src/algorithms/generate.rs
```diff
@@ -162,5 +163,6 @@ mod tests {
 
     key_generation!(key_generation_multi_5_64, 5, 64);
     key_generation!(key_generation_multi_8_576, 8, 576);
-    key_generation!(key_generation_multi_16_1024, 16, 1024);
+    // TODO: reenable, currently slow
+    // key_generation!(key_generation_multi_16_1024, 16, 1024);
```
tarcieri, line 167: 😢

src/algorithms/rsa.rs
```diff
@@ -34,97 +33,86 @@ pub fn rsa_encrypt<K: PublicKeyParts>(key: &K, m: &BigUint) -> Result<BigUint> {
...
+        // c^d (mod n)
+        pow_mod_params(&c, d, n_params)
+    } else {
+        // We have the precalculated values needed for the CRT.
+
+        let dp = priv_key.dp().expect("precomputed");
+        let dq = priv_key.dq().expect("precomputed");
+        let qinv = priv_key.qinv().expect("precomputed");
+        let p_params = priv_key.p_params().expect("precomputed");
+        let q_params = priv_key.q_params().expect("precomputed");
```
tarcieri, line 71: The previous pattern-based approach seemed a little more elegant than this

src/algorithms/rsa.rs
```diff
@@ -154,126 +142,179 @@ pub fn rsa_decrypt_and_check<R: CryptoRngCore + ?Sized>(
...
     if !sqrt_is_whole_number {
         return Err(Error::InvalidArguments);
     }
-    let p = (&*b + &*y) / &two;
-    let q = (&*b - &*y) / two;
+
+    let bits = core::cmp::max(b.bits_precision(), y.bits_precision());
+    let two = NonZero::new(BoxedUint::from(2u64))
+        .expect("2 is non zero")
+        .widen(bits);
```
tarcieri, line 278: Something which makes this pattern better would be nice. We have various `NonZero::new_unwrap` constructors which allow you to infallibly construct a `NonZero`.

src/algorithms/generate.rs
tarcieri: @dignifiedquire so if I understand correctly, the purpose of this calculation is ultimately for a check that keygen terminates in a reasonable amount of time?

Cargo.toml
fjarri: If all you need from there is `logf(x)`, you can probably just approximate it with `x.bits() / log2(e)`, pre-calculating `log2(e)`.

fjarri: Or even take `log2(e) == 1` if the approximation allows it :)

src/algorithms/generate.rs
```diff
@@ -93,14 +95,14 @@ pub(crate) fn generate_multi_prime_key_with_exp<R: CryptoRngCore + ?Sized>(
 
         let n = compute_modulus(&primes);
 
-        if n.bits() != bit_size {
+        if n.bits() as usize != bit_size {
             // This should never happen for nprimes == 2 because
             // gen_prime should set the top two bits in each prime.
```
fjarri, line 100: The comment should be updated for the new name of the function.

Also I don't quite understand "gen_prime should set the top two bits in each prime" - what was that behavior for? `generate_prime_with_rng()` only sets one top bit, to ensure that the returned primes are of the requested bit size.

tarcieri: I was a bit curious about that myself. I saw @FiloSottile mention the same thing in this post:

https://words.filippo.io/dispatches/rsa-keygen-bench/

> There is almost nothing special to selecting prime candidates. You draw an appropriately sized random number from a CSPRNG, and to avoid wasting time, you set the least significant bit and the two most significant bits: large even numbers are not prime, and setting the top two guarantees N won’t come out too small.

src/encoding.rs
```diff
@@ -51,13 +52,32 @@ impl TryFrom<pkcs8::PrivateKeyInfoRef<'_>> for RsaPrivateKey {
...
+            .map_err(|_| key_malformed)?;
+        let e = Option::from(e).ok_or(key_malformed)?;
+
+        let d = BoxedUint::from_be_slice(pkcs1_key.private_exponent.as_bytes(), bits)
+            .map_err(|_| key_malformed)?;
+
+        let prime1 = BoxedUint::from_be_slice(pkcs1_key.prime1.as_bytes(), bits)
+            .map_err(|_| key_malformed)?;
+        let prime2 = BoxedUint::from_be_slice(pkcs1_key.prime2.as_bytes(), bits)
+            .map_err(|_| key_malformed)?;
```
tarcieri, line 77: I'm not sure I quote understand why these conversions got more convoluted but perhaps there are upstream APIs which could be added to assist?

src/algorithms/generate.rs
FiloSottile: If you set only the top bit of a n/2 bits number, you get a value >= 2^(n/2-1). If you multiply two of them, you get a value >= 2^(n-2), which is at a minimum n-1 bits long, one too short for a n-bit modulus.

fjarri: Ah I see, that makes sense. Do you think it should be the default behavior in `crypto-primes`?

FiloSottile: If you mainly need the primes for RSA, yes, you are only wasting CPU rejecting small moduli after you found the primes. If you're using the primes for something else, it might depend on what they are for.

src/encoding.rs
richarddd: Moving this to a method would clean it up a bunch. Call out to it and map_err from the call

src/algorithms/generate.rs
dignifiedquire: @fjarri are you planning on changing the behaviour?

Cargo.toml
dignifiedquire: `bits()` seems to not be around either..

src/encoding.rs
dignifiedquire: cleaned up a bit

src/algorithms/rsa.rs
dignifiedquire: fixed

src/lib.rs
dignifiedquire: added

README.md
dignifiedquire: fixed all around

src/algorithms/rsa.rs
dignifiedquire:
```
278 |     let two = NonZero::new_unwrap(2).widen(bits);
    |                        ^^^^^^^^^^ multiple `new_unwrap` found
    |
    = note: candidate #1 is defined in an impl for the type `crypto_bigint::NonZero<Limb>`
    = note: candidate #2 is defined in an impl for the type `crypto_bigint::NonZero<crypto_bigint::Uint<LIMBS>>`
```

seems to not exist for the`BoxedUint` version

src/algorithms/generate.rs
fjarri: I thought about it, and I don't think it should be the default - too specialized. It can be achieved with custom sieves, but I need to make a release first

dignifiedquire: that's fair, please do let me know if you have a release/how to make replicate the desired behavior with custom sieves

tarcieri:
FYI: release PR for `crypto-bigint` v0.6.0 is up: https://github.com/RustCrypto/crypto-bigint/pull/750

src/algorithms/generate.rs
```diff
@@ -109,30 +108,41 @@ pub(crate) fn generate_multi_prime_key_with_exp<R: CryptoRngCore + ?Sized>(
...
     Ok(RsaPrivateKeyComponents {
         n: n_final,
-        e: exp.clone(),
+        e: exp,
         d: d_final,
         primes,
     })
 }
 
+/// Natural logrithm for `f64`.
```
YgorSouza, line 117: Typo here: `logrithm -> logarithm` (and the same on line 123). Could add the [typos](https://github.com/marketplace/actions/typos-action) GitHub action to the repo to avoid that.

src/algorithms/rsa.rs
tarcieri: @dignifiedquire aah, yeah the idea would be to const construct a `NonZero<Uint>` first and either convert it to a `NonZero<BoxedUint>` first or make it possible to use `NonZero<Uint>` as an operand.

src/algorithms/generate.rs
fjarri: Opened https://github.com/entropyxyz/crypto-primes/pull/69. If that is merged, you will need to use `sieve_and_find()` to access this functionality.

src/algorithms/generate.rs
kjvalencik: For what it's worth, most other implementations store it as a small number, so I'm not sure it's _practically_ possible to have a large exponent.

For example, [ring stores it as a `u64`](https://github.com/briansmith/ring/blob/ba96d3d078e26882d340ad5ba5b22e21b97fcc70/src/rsa/public_exponent.rs#L7) and [golang uses an `int`](https://cs.opensource.google/go/go/+/master:src/crypto/rsa/rsa.go;l=70).

tarcieri (approved):
I could continue scanning this PR looking for improvements, but I don't see any worth blocking a merge on. I can retroactively file some issues perhaps.

src/algorithms/generate.rs
tarcieri: Nice! Thanks for `SetBits::TwoMsb`

src/algorithms/generate.rs
dignifiedquire: fixed, thanks

src/algorithms/generate.rs
dignifiedquire: done, thank you

dignifiedquire:
@tarcieri updated to the latest version, and fixed a subtle padding bug that I found, please take a final look before I merge

Cargo.toml
fjarri: Well, the only instance of `logf()` usage is
```
let prime_limit = (1u64 << (bit_size / nprimes) as u64) as f64;
let mut pi = prime_limit / (logf(prime_limit) - 1f64);
```
So you can just do
```
let log2_e = 1.4426950408889634;
let mut pi = prime_limit / ((bit_size / nprimes) / log2_e  - 1); // with whatever type conversions required
```
Does that work?

dignifiedquire: I couldn't get it to work, but I did find a nice and short ln implementation on the internet that I added. and now no more `libm` dependency

src/algorithms/generate.rs
```diff
@@ -186,4 +200,19 @@ mod tests {
...
+        let mut rng = ChaCha8Rng::from_seed([42; 32]);
+
+        for i in 0..100 {
+            println!("round {i}");
+            let prime_limit: f64 = rng.gen();
+            let a = logf(prime_limit);
+            let b = logf_approx(prime_limit as f32);
+
+            let diff = a - b as f64;
+            assert!(diff < 0.001, "{} != {}", a, b);
```
praseodym, line 215: What is this supposed to test? `logf` calls `logf_approx`, so this currently tests whether casting from `f32` to `f64` does not result in big difference?

praseodym: Maybe test against [`ln`](https://doc.rust-lang.org/std/primitive.f64.html#method.ln) from the standard library instead?

dignifiedquire: logf calls ln when compiled under std

Cargo.toml
dignifiedquire: the results are not even close, they are completely different numbers

fjarri: That's pretty strange. Seems to work for me:
```rust
fn main() {
    let bit_size = 1024;
    let nprimes = 42;

    let prime_limit = (1u64 << (bit_size / nprimes) as u64) as f64;
    let pi_ref = prime_limit / (prime_limit.ln() - 1f64);

    println!("{pi_ref}");

    let log2_e = 1.4426950408889634;
    let pi_test = prime_limit / ((bit_size / nprimes) as f64 / log2_e  - 1f64); // with whatever type conversions required

    println!("{pi_test}");
}
```
```
1073018.5351041532
1073018.5351041532
```

src/algorithms/generate.rs
praseodym: I missed that from just reading the diff, the test makes sense now!

src/key.rs
```diff
@@ -461,73 +529,85 @@ impl RsaPrivateKey {
...
+
+    let n = match n.bits_precision().cmp(&bits_precision) {
+        Ordering::Less => n.widen(bits_precision),
+        Ordering::Equal => n.clone(),
+        Ordering::Greater => n.shorten(bits_precision),
+    };
+
+    let n_reduced = n.rem_vartime(&modulus).widen(p.bits_precision());
+    BoxedMontyForm::new(n_reduced, p.clone())
+}
```
tarcieri, line 609: Am I missing where this is actually used?

Perhaps it should be named `reduce_vartime` since it uses `rem_vartime`?

src/pkcs1v15.rs
```diff
@@ -145,7 +143,10 @@ fn encrypt<R: CryptoRngCore + ?Sized>(
     key::check_public(pub_key)?;
 
     let em = pkcs1v15_encrypt_pad(rng, msg, pub_key.size())?;
-    let int = Zeroizing::new(BigUint::from_bytes_be(&em));
+    let int = BoxedUint::from_be_slice(
+        &em,
+        crate::traits::keys::PublicKeyParts::n_bits_precision(pub_key),
```
tarcieri, line 148: Is there some reason you aren't just importing this, especially since it's used repeatedly? So it could just be...
```suggestion
        PublicKeyParts::n_bits_precision(pub_key),
```

src/pkcs1v15/signature.rs
```diff
@@ -15,10 +15,9 @@ use spki::{
 /// `RSASSA-PKCS1-v1_5` signatures as described in [RFC8017 § 8.2].
 ///
 /// [RFC8017 § 8.2]: https://datatracker.ietf.org/doc/html/rfc8017#section-8.2
-#[derive(Clone, PartialEq, Eq)]
+#[derive(Debug, Clone, PartialEq, Eq)]
 pub struct Signature {
-    pub(super) inner: BigUint,
-    pub(super) len: usize,
+    pub(super) inner: BoxedUint,
```
tarcieri, line 20: Nice, fixed (dynamic) precision saves the day!

src/pkcs1v15/signing_key.rs
```diff
@@ -49,7 +49,7 @@ where
     }
 
     /// Generate a new signing key with a prefix for the digest `D`.
-    pub fn random<R: CryptoRngCore + ?Sized>(rng: &mut R, bit_size: usize) -> Result<Self> {
+    pub fn random<R: CryptoRngCore>(rng: &mut R, bit_size: usize) -> Result<Self> {
```
tarcieri, line 52: Is there a particular reason you're getting rid of `?Sized` on these? It makes it possible to use a `dyn CryptoRngCore`

src/pss.rs
```diff
@@ -205,12 +205,15 @@ fn sign_pss_with_salt<T: CryptoRngCore>(
...
 
-    uint_to_zeroizing_be_pad(
-        rsa_decrypt_and_check(priv_key, blind_rng, &BigUint::from_bytes_be(&em))?,
-        priv_key.size(),
-    )
+    let em = emsa_pss_encode(hashed, em_bits as _, salt, digest)?;
+
+    let em = BoxedUint::from_be_slice(
+        &em,
+        crate::traits::keys::PublicKeyParts::n_bits_precision(priv_key),
```
tarcieri, line 213: Similar concern re: `PublicKeyParts`

src/pss/signature.rs
```diff
@@ -35,26 +34,22 @@ impl TryFrom<&[u8]> for Signature {
...
-            inner: BigUint::from_bytes_be(bytes),
-        })
+        let len = bytes.len();
+        let inner = BoxedUint::from_be_slice(bytes, len as u32 * 8);
+
+        #[cfg(feature = "std")]
+        let inner = inner
+            .map_err(|e| Box::new(e) as Box<dyn core::error::Error + Send + Sync + 'static>)?;
+        #[cfg(not(feature = "std"))]
+        let inner = inner.map_err(|_| signature::Error::new())?;
```
tarcieri, line 44: Ugh, should really fix this upstream in the `signature` crate

tarcieri (approved):
Various little nits but things are looking a lot better on this pass

Cargo.toml
Fethbita: @dignifiedquire

tarcieri:
@dignifiedquire can we land this?

fjarri:
There's a minor thing of needing the `libm` dependency (https://github.com/RustCrypto/RSA/pull/394#discussion_r1903142954), but I can make a separate PR for that

fjarri:
Also the 6-10x performance drop is quite surprising, but that can be addressed in a follow-up as well.

dignifiedquire:
> There's a minor thing of needing the `libm` dependency ([#394 (comment)](https://github.com/RustCrypto/RSA/pull/394#discussion_r1903142954)), but I can make a separate PR for that

I removed that, there is some other follow ups from you and @tarcieri I wanted to get to, but we can also merge as is

src/pkcs1v15/signing_key.rs
dignifiedquire: > Is there a particular reason you're getting rid of ?Sized on these? It makes it possible to use a dyn CryptoRngCore

I don't think so, can revert probably

dignifiedquire: I checked, and I removed it because I can't pass that constraint to `crypto_primes`

src/pkcs1v15.rs
dignifiedquire: cleaned these up

src/key.rs
dignifiedquire: renamed, and moved it ot where it is used

dignifiedquire:
Thank you everyone who helped make this a reality!  And sorry for this taking so long..

src/pkcs1v15/signing_key.rs
fjarri: Adding it to `crypto-primes` is unfortunately blocked by `crypto-bigint` not supporting it.

fjarri: Made https://github.com/RustCrypto/crypto-bigint/pull/760 for that

PR #400, opened 2023-12-10
Marvin toolkit container

xuganyu96:
This is an attempt to make the Marvin attack analysis more reproducible for `RustCrypto/RSA`. The Dockerfile and entrypoint script are both rough proofs of concept. It would be great if the maintainers can suggest how we can handle the outputs of the analysis, whether to test `rsa1024` and `rsa4096`, whether this should be part of CI/CD, among other things.

### DIFF
--- a/marvin-toolkit/Cargo.toml
+++ b/marvin-toolkit/Cargo.toml
@@ -9,5 +9,5 @@ clap = { version = "4", features = ["derive"] }
 rsa = "0.9"
 
 [patch.crates-io]
-rsa = { git = "https://github.com/xuganyu96/RSA", branch = "const-crypto-biguint" }
+rsa = { git = "https://github.com/RustCrypto/RSA", branch = "const-crypto-biguint" }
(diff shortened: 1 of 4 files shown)


tarcieri:
This looks fine to me. Will try it out locally.

xuganyu96:
@tarcieri @dignifiedquire 

This PR is ready for review. The main feature of this PR is a container that packages the Marvin toolkit. See the added README.md for how to build and run the container. Some notable features include:

1. CLI allows specifying `-s <RSA_bits> -n <repeat>`
2. Custom `Cargo.toml` for the test harness if custom version of `RustCrypto/RSA` and/or `RustCrypto/crypto-bigint` is needed
3. Preserve the generated keys, ciphertexts, and analysis results by copying them into a location that can be mounted into the host file system

tarcieri:
Cool, I'll try to test it out soon

tarcieri (approved):

PR #403, opened 2024-01-19
Bump version to v0.10.0-pre

tarcieri:
NOTE: not for release.

This signifies that we are going to make breaking changes to the `master` branch which are incompatible with v0.9 releases.

The first prerelease of this series published to crates.io will be v0.10.0-pre.0 at some point in the future.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "rsa"
-version = "0.9.6"
+version = "0.10.0-pre"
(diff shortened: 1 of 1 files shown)


tarcieri:
I need to bump the format/signature dependencies to be able to upgrade `x509-cert`, which I'll do in an immediate followup after this

tarcieri:
Seems we need #404 first

Issue #419, opened 2024-03-18
Serde support for `rsa::pkcs1v15::Signature`


LWEdslev:
Currently `rsa::pkcs1v15::Signature` does not support serde Serialization and Deserialization. 
Is there any good explanation for this?

tarcieri:
It would be good to add, yes

PR #420, opened 2024-03-18
added serdect support

LWEdslev:
Added serde support (Serialization, Deserialization) for structs in `src/pkcs1v15/`:
```
DecryptingKey
EncryptingKey
Signature
SigningKey<D>
VerifyingKey<D>
```
#419


tarcieri:
Since all of these have a straightforward serialization to bytes, it would be good to use that, rather than custom derive.

@dignifiedquire perhaps we should use the `serdect` crate for this like in e.g. https://github.com/rustcrypto/elliptic-curves ?

LWEdslev:
> Since all of these have a straightforward serialization to bytes, it would be good to use that, rather than custom derive.
> 
> @dignifiedquire perhaps we should use the `serdect` crate for this like in e.g. https://github.com/rustcrypto/elliptic-curves ? 

Why has `serdect` not been used in `src/key.rs PrivateKey`?

tarcieri:
It hasn't been used in the `rsa` crate at all yet, but has opinions around how data is serialized, namely it uses strategies which are constant time across various Serde serializers, regardless of if they're for binary or human readable formats (where hex is used to encode the latter), which is important when serializing private keys.

LWEdslev:
> perhaps we should use the `serdect`

So kinda like this?
```Rust
impl Serialize for Signature {
    fn serialize<S>(&self, serializer: S) -> core::result::Result<S::Ok, S::Error>
    where
        S: serdect::serde::Serializer,
    {
        serdect::slice::serialize_hex_lower_or_bin(&self.inner.to_bytes_be(), serializer)
    }
}

impl<'de> Deserialize<'de> for Signature {
    fn deserialize<D>(deserializer: D) -> core::result::Result<Self, D::Error>
    where
        D: serdect::serde::Deserializer<'de>,
    {
        let bytes = serdect::slice::deserialize_hex_or_bin_vec(deserializer)?;
        let inner = BigUint::from_bytes_be(&bytes);

        Ok(Self {
            inner,
            len: bytes.len(),
        })
    }
}
```
I can also do the others

dignifiedquire:
sure 👍

LWEdslev:
I guess only `Signature` has a straight forward `serdect` implementation, since the other structs rely on `RsaPublicKey`, `RsaPrivateKey` and more, so that's gonna have to do for now.

tarcieri:
@LWEdslev the `elliptic-curve` crates use the SPKI serialization for public keys. You could do something similar for `RsaPublicKey` and `VerifyingKey` at least: computing the SPKI serialization with [`EncodePublicKey::to_public_key_der`](https://docs.rs/pkcs8/latest/pkcs8/trait.EncodePublicKey.html#tymethod.to_public_key_der), and then encoding that using `serdect`.

LWEdslev:
@tarcieri 
I see, I'll give it a whirl

LWEdslev:
`RsaPublicKey`, `RsaPrivateKey` and the structs in `pkcs1v15` now have the `serdect` implementations

Cargo.toml
```diff
@@ -25,6 +25,7 @@ pkcs8 = { version = "=0.11.0-pre.0", default-features = false, features = ["allo
 signature = { version = "=2.3.0-pre.3", default-features = false , features = ["alloc", "digest", "rand_core"] }
 spki = { version = "=0.8.0-pre.0", default-features = false, features = ["alloc"] }
 zeroize = { version = "1.5", features = ["alloc"] }
+serdect = "0.2.0"
 
 # optional dependencies
 sha1 = { version = "=0.11.0-pre.3", optional = true, default-features = false, features = ["oid"] }
```
tarcieri, line 31:
```suggestion

# optional dependencies
sha1 = { version = "=0.11.0-pre.3", optional = true, default-features = false, features = ["oid"] }
serdect = { version = "0.2.0", optional = true }
```

Then add it to the `serde` feature below as `dep:serdect` alongside the existing `dep:serde`

LWEdslev:
Should I add some `serde_test` tests like the `key::tests::test_serde` test for the new implementations?

tarcieri:
Sure

LWEdslev:
Due to the `PhantomData<D>` fields that has `D: Digest`, I can't use `PartialEq` on all the structs and therefore I can't use `serde_test::assert_tokens`. I could not find a better solution than using `serde_json` for those structs.

tarcieri:
Can you post the code that isn’t working?

LWEdslev:
> Can you post the code that isn’t working?

For the `oaep::EncryptingKey` which uses the `D: Digest` generic I tried something like this:
```Rust
#[test]
    #[cfg(feature = "serde")]
    fn test_serde() {
        use super::*;
        use rand_chacha::{rand_core::SeedableRng, ChaCha8Rng};
        use serde_test::{assert_tokens, Configure, Token};

        let mut rng = ChaCha8Rng::from_seed([42; 32]);
        let priv_key = crate::RsaPrivateKey::new(&mut rng, 64).expect("failed to generate key");
        let encrypting_key = EncryptingKey::<sha2::Sha256>::new(priv_key.to_public_key());

        let tokens = [
            Token::Struct { name: "EncryptingKey", len: 1 },
            Token::Str("inner"),
            Token::Str("3024300d06092a864886f70d01010105000313003010020900cc6c6130e35b46bf0203010001"),
            Token::Str("label"),
            Token::Str("None"),
            Token::StructEnd,
        ];
        assert_tokens(&encrypting_key.readable(), &tokens);
    }
```

To use the `assert_tokens` I need to add `PartialEq` to the `EcnryptingKey` struct. Then I get an error when running `cargo test --features serde`:
```Rust
error[E0277]: can't compare `CoreWrapper<CtVariableCoreWrapper<Sha256VarCore, UInt<UInt<UInt<UInt<UInt<UInt<UTerm, B1>, B0>, B0>, B0>, B0>, B0>, OidSha256>>` with `CoreWrapper<CtVariableCoreWrapper<Sha256VarCore, UInt<UInt<UInt<UInt<UInt<UInt<UTerm, B1>, B0>, B0>, B0>, B0>, B0>, OidSha256>>`
```

src/pkcs1v15/encrypting_key.rs
```diff
@@ -1,12 +1,15 @@
...
 use alloc::vec::Vec;
+#[cfg(feature = "serde")]
+use serde::{Deserialize, Serialize};
 use rand_core::CryptoRngCore;
 
 /// Encryption key for PKCS#1 v1.5 encryption as described in [RFC8017 § 7.2].
 ///
 /// [RFC8017 § 7.2]: https://datatracker.ietf.org/doc/html/rfc8017#section-7.2
-#[derive(Debug, Clone)]
+#[derive(Debug, Clone, PartialEq)]
```
tarcieri, line 11: The problem is you're trying to derive `PartialEq` which is getting tripped up on the phantom data parameter so you can't do that.

Instead handwrite the impl instead.

LWEdslev:
Now all of them have the `serde_test` tests

LWEdslev:
So what's next?

src/pkcs1v15/signing_key.rs
```diff
@@ -253,3 +260,60 @@ where
...
+#[cfg(feature = "serde")]
+impl<D> Serialize for SigningKey<D>
+where
+    D: Digest,
+{
+    fn serialize<S>(&self, serializer: S) -> core::result::Result<S::Ok, S::Error>
+    where
+        S: serdect::serde::Serializer,
+    {
+        let der = self.inner.to_pkcs8_der().map_err(ser::Error::custom)?;
```
tarcieri, line 279:
```suggestion
        let der = self.to_pkcs8_der().map_err(ser::Error::custom)?;
```

src/pkcs1v15/signing_key.rs
```diff
@@ -253,3 +260,60 @@ where
...
+where
+    D: Digest + AssociatedOid,
+{
+    fn deserialize<De>(deserializer: De) -> core::result::Result<Self, De::Error>
+    where
+        De: serdect::serde::Deserializer<'de>,
+    {
+        let der_bytes = serdect::slice::deserialize_hex_or_bin_vec(deserializer)?;
+        let pki = PrivateKeyInfo::from_der(&der_bytes).map_err(de::Error::custom)?;
+        Self::try_from(pki).map_err(de::Error::custom)
```
tarcieri, line 295:
```suggestion
        Self::from_pkcs8_der(&der_bytes).map_err(de::Error::custom)
```

src/pkcs1v15/verifying_key.rs
```diff
@@ -201,3 +210,64 @@ where
...
+#[cfg(feature = "serde")]
+impl<D> Serialize for VerifyingKey<D>
+where
+    D: Digest,
+{
+    fn serialize<S>(&self, serializer: S) -> Result<S::Ok, S::Error>
+    where
+        S: serde::Serializer,
+    {
+        let der = self.inner.to_public_key_der().map_err(ser::Error::custom)?;
```
tarcieri, line 232:
```suggestion
        let der = self.to_public_key_der().map_err(ser::Error::custom)?;
```

src/pkcs1v15/verifying_key.rs
```diff
@@ -201,3 +210,64 @@ where
...
+where
+    D: Digest + AssociatedOid,
+{
+    fn deserialize<De>(deserializer: De) -> Result<Self, De::Error>
+    where
+        De: serde::Deserializer<'de>,
+    {
+        let der_bytes = serdect::slice::deserialize_hex_or_bin_vec(deserializer)?;
+        let spki = SubjectPublicKeyInfo::from_der(&der_bytes).map_err(de::Error::custom)?;
+        Self::try_from(spki).map_err(de::Error::custom)
```
tarcieri, line 248:
```suggestion
        Self::from_public_key_der(&der_bytes).map_err(de::Error::custom)
```

src/pss/blinded_signing_key.rs
```diff
@@ -198,3 +203,65 @@ where
...
+#[cfg(feature = "serde")]
+impl<D> Serialize for BlindedSigningKey<D>
+where
+    D: Digest,
+{
+    fn serialize<S>(&self, serializer: S) -> core::result::Result<S::Ok, S::Error>
+    where
+        S: serde::Serializer,
+    {
+        let der = self.inner.to_pkcs8_der().map_err(ser::Error::custom)?;
```
tarcieri, line 225:
```suggestion
        let der = self.to_pkcs8_der().map_err(ser::Error::custom)?;
```

src/pss/blinded_signing_key.rs
```diff
@@ -198,3 +203,65 @@ where
...
+{
+    fn deserialize<De>(deserializer: De) -> core::result::Result<Self, De::Error>
+    where
+        De: serde::Deserializer<'de>,
+    {
+        let der_bytes = serdect::slice::deserialize_hex_or_bin_vec(deserializer)?;
+        let pki = PrivateKeyInfo::from_der(&der_bytes).map_err(de::Error::custom)?;
+        RsaPrivateKey::try_from(pki)
+            .map_err(de::Error::custom)
+            .map(Self::new)
```
tarcieri, line 243:
```suggestion
        Self::from_pkcs8_der(&der_bytes).map_err(de::Error::custom)
```

src/pss/signature.rs
```diff
@@ -73,3 +75,48 @@ impl Display for Signature {
...
     }
 }
+
+#[cfg(feature = "serde")]
+impl Serialize for Signature {
+    fn serialize<S>(&self, serializer: S) -> core::result::Result<S::Ok, S::Error>
+    where
+        S: serdect::serde::Serializer,
+    {
+        serdect::slice::serialize_hex_lower_or_bin(&self.inner.to_bytes_be(), serializer)
```
tarcieri, line 85:
```suggestion
        serdect::slice::serialize_hex_lower_or_bin(&self.to_bytes(), serializer)
```

src/pss/signature.rs
```diff
@@ -73,3 +75,48 @@ impl Display for Signature {
...
+    where
+        D: serdect::serde::Deserializer<'de>,
+    {
+        let bytes = serdect::slice::deserialize_hex_or_bin_vec(deserializer)?;
+        let inner = BigUint::from_bytes_be(&bytes);
+
+        Ok(Self {
+            inner,
+            len: bytes.len(),
+        })
```
tarcieri, line 101:
```suggestion
        serdect::slice::deserialize_hex_or_bin_vec(deserializer)?.as_slice().try_into()
```

src/pss/signing_key.rs
```diff
@@ -220,3 +226,65 @@ where
...
+#[cfg(feature = "serde")]
+impl<D> Serialize for SigningKey<D>
+where
+    D: Digest,
+{
+    fn serialize<S>(&self, serializer: S) -> core::result::Result<S::Ok, S::Error>
+    where
+        S: serdect::serde::Serializer,
+    {
+        let der = self.inner.to_pkcs8_der().map_err(ser::Error::custom)?;
```
tarcieri, line 248:
```suggestion
        let der = self.to_pkcs8_der().map_err(ser::Error::custom)?;
```

src/pss/signing_key.rs
```diff
@@ -220,3 +226,65 @@ where
...
+    fn deserialize<De>(deserializer: De) -> core::result::Result<Self, De::Error>
+    where
+        De: serdect::serde::Deserializer<'de>,
+    {
+        let der_bytes = serdect::slice::deserialize_hex_or_bin_vec(deserializer)?;
+        let pki = PrivateKeyInfo::from_der(&der_bytes).map_err(de::Error::custom)?;
+        RsaPrivateKey::try_from(pki)
+            .map_err(de::Error::custom)
+            .map(Self::new)
+    }
```
tarcieri, line 267:
```suggestion
        Self::from_pkcs8_der(&der_bytes).map_err(de::Error::custom)
```

src/pss/verifying_key.rs
```diff
@@ -156,3 +162,66 @@ where
...
+#[cfg(feature = "serde")]
+impl<D> Serialize for VerifyingKey<D>
+where
+    D: Digest,
+{
+    fn serialize<S>(&self, serializer: S) -> Result<S::Ok, S::Error>
+    where
+        S: serde::Serializer,
+    {
+        let der = self.inner.to_public_key_der().map_err(ser::Error::custom)?;
```
tarcieri, line 184:
```suggestion
        let der = self.to_public_key_der().map_err(ser::Error::custom)?;
```

src/pss/verifying_key.rs
```diff
@@ -156,3 +162,66 @@ where
...
+    fn deserialize<De>(deserializer: De) -> Result<Self, De::Error>
+    where
+        De: serde::Deserializer<'de>,
+    {
+        let der_bytes = serdect::slice::deserialize_hex_or_bin_vec(deserializer)?;
+        let spki = SubjectPublicKeyInfo::from_der(&der_bytes).map_err(de::Error::custom)?;
+        RsaPublicKey::try_from(spki)
+            .map_err(de::Error::custom)
+            .map(Self::new)
+    }
```
tarcieri, line 203:
```suggestion
        Self::from_public_key_der(&der_bytes).map_err(de::Error::custom)
```

tarcieri:
As a bit of a meta comment on the above suggestions: all of these types have built-in serialization support that the `serde` impls can be thin wrappers for, rather than duplicating. It shouldn't be necessary to reach into `inner`, for example

LWEdslev:
Yeah, that's better

LWEdslev:
How do I make the `pss::SigningKey` use the correct algorithm OID when decoding? 
`EncodePrivateKey` uses `pkcs1::ALGORITHM_ID` but `ID_RSASSA_PSS` is required. 
I could reimplement `EncodePrivateKey for pss::SigningKey` where it uses `ID_RSASSA_PSS` from the bottom with just that one change but that would cause code duplication since it would be almost identical with the `EncodePrivateKey for RsaPrivateKey`, so maybe extract the common behavior or is there a simple obvious way?

tarcieri:
@LWEdslev either OID should work now (`pkcs1::ALGORITHM_ID` or `ID_RSASSA_PSS`), now that #424 is merged, though you may need to rebase

LWEdslev:
@tarcieri I don't see how either OID should work for the `pss::SigningKey`. 
With the merge of #424 the `TryFrom<pkcs8::PrivateKeyInfo<'_>> for SigningKey<D>` asserts 
```Rust
private_key_info
            .algorithm
            .assert_algorithm_oid(ID_RSASSA_PSS)?;
```
`ID_RSASSA_PSS` is the OID `1.2.840.113549.1.1.10`
but the encoding is done with
```Rust
self.inner.to_pkcs8_der()
```
And the inner `RsaPrivateKey` has `EncodePrivateKey` implemented such that algorithms (`pkcs1::ALGORITHM_ID`) OID is `1.2.840.113549.1.1.1` 
So encoding a `pss::SigningKey` it can't be decoded back into itself.

tarcieri:
@LWEdslev aah, so it does. It should probably use `verify_algorithm_id` which accepts either

LWEdslev:
@tarcieri Will you review this when you have time?

tarcieri:
Yes

src/pss/signing_key.rs
```diff
@@ -227,11 +232,68 @@ where
...
 }
 
 impl<D> ZeroizeOnDrop for SigningKey<D> where D: Digest {}
+
+impl<D> PartialEq for SigningKey<D>
+where
+    D: Digest,
+{
+    fn eq(&self, other: &Self) -> bool {
+        self.inner == other.inner && self.salt_len == other.salt_len
```
tarcieri, line 247: Hmm, I just noticed the derived `Eq`/`PartialEq` on `RsaPrivateKey` are not constant-time, which seems bad, although not directly related to this PR except it's indirectly invoking it.

src/pss/signing_key.rs
```diff
@@ -227,11 +232,68 @@ where
     type Error = pkcs8::Error;
 
     fn try_from(private_key_info: pkcs8::PrivateKeyInfo<'_>) -> pkcs8::Result<Self> {
-        private_key_info
-            .algorithm
-            .assert_algorithm_oid(ID_RSASSA_PSS)?;
+        verify_algorithm_id(&private_key_info.algorithm)?;
```
tarcieri, line 235: I guess this is a fix for handling PKCS#8 keys with PSS OIDs

tarcieri (approved):

Issue #422, opened 2024-03-26
Is RSA PSS key loading from PKCS#8 supported?


taleks:
Hi,

Are RSA PSS keys supported in `rsa` crate for loading from `PKCS#8`? Documentation has that small greyish checkbox for "PSS: Sign & Verify" bullet (though it is not clear what does it mean as there is no any legend) plus https://docs.rs/rsa/latest/rsa/index.html#pkcs8-rsa-key-encoding mentions related traits, so I assumed it is supported. But maybe I was a bit too hasty in that  assumption. 

If those are supported what is the right way to load RSA PSS private key?

### Context

I am trying to load private key https://www.rfc-editor.org/rfc/rfc9421.html#name-example-rsa-pss-key

Test code is:

```rust
use base64::Engine;
use base64::prelude::BASE64_STANDARD;
use pkcs8::{DecodePrivateKey, PrivateKeyInfo};
use rsa::RsaPrivateKey;

const RSA_PSS: &str = include_str!("data/rsa2048_rfc9421.pem");

fn main() {
    let line = RSA_PSS.replace("\n", "")
        .replace("-----BEGIN PRIVATE KEY-----", "")
        .replace("-----END PRIVATE KEY-----", "");

    let bytes = BASE64_STANDARD.decode(line).unwrap();
    let pk = PrivateKeyInfo::try_from(bytes.as_slice()).unwrap();

    println!("Key info: {:?}", pk);

    if let Err(err) = RsaPrivateKey::from_pkcs8_pem(RSA_PSS) {
        println!("Error: {err}")
    }
}
```

Output is:

```
Key info: PrivateKeyInfo { version: V1, algorithm: AlgorithmIdentifier { oid: ObjectIdentifier(1.2.840.113549.1.1.10), parameters: None }, public_key: None, .. }
Error: public key error: unknown/unsupported algorithm OID: 1.2.840.113549.1.1.1
```

My understanding is that `.10` is OID for `rsassa-pss`, while `.1` is `rsaEncryption`.

Validation fails here https://github.com/RustCrypto/RSA/blob/master/src/encoding.rs#L15-L23

`pkcs1::ALGORITHM_OID` comes from https://github.com/RustCrypto/formats/blob/master/pkcs1/src/lib.rs#L55 where it is defined as `1.2.840.113549.1.1.1`.

### Not really related but...

[...]

PR #424, opened 2024-03-27
pss: support DecodePrivateKey and DecodePublicKey traits

lumag:
Implement necessary conversion traits to enable DecodePrivateKey and DecodePublicKey traits implementation.


tarcieri (approved):
Approved, although I'm curious if these keys support anything besides just an alternative  OID. Anyway, this should be sufficient to close #422.

tarcieri:
It seems the OID is from [RFC7427](https://datatracker.ietf.org/doc/html/rfc7427) but it doesn't say anything about its usage in PKCS#8. Anyway, merging.

PR #425, opened 2024-04-02
Stop gap fix serde

pinkforest:
Cc/ @dignifiedquire @tarcieri 

serde skip requires a default for `BoxedMontyParams` associated n_params since it's not an option at RsaPublicKey

### DIFF
--- a/src/key.rs
+++ b/src/key.rs
@@ -19,6 +19,13 @@ use crate::errors::{Error, Result};
 use crate::traits::keys::{CrtValue, PrivateKeyParts, PublicKeyParts};
 use crate::traits::{PaddingScheme, SignatureScheme};
 
+// Serde requires default upon skippin non-Option field
+// TODO: stop gap for feature serde compile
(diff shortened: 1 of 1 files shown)


src/key.rs
```diff
@@ -31,7 +38,7 @@ pub struct RsaPublicKey {
     /// Typically 0x10001 (65537)
     e: u64,
 
-    #[cfg_attr(feature = "serde", serde(skip))]
+    #[cfg_attr(feature = "serde", serde(skip, default = "default_nparams"))]
```
dignifiedquire, line 41: is there no way to construct this during deserialization from the actual values?

pinkforest: yeah would need to impl 'De obvs - just needed to get this to compile in the meantime to see errors w/ all-featuers  - will plug this later once I understand more what I'm doing in here :rofl:

dignifiedquire:
not needed anymore

PR #447, opened 2024-08-20
Update dependencies

Snowiiii:
Hey, Just a some regularly dependency updates


baloo:
Dependabot should already take care of that. Also the Cargo.lock was mostly up to date already.

The patch version doesn't really matter cargo will always pick the latest (see https://doc.rust-lang.org/cargo/reference/specifying-dependencies.html#specifying-dependencies-from-cratesio). This should only be required to bump the patch version of `Cargo.toml` to fix a regression / security issue.

PR #448, opened 2024-09-14
pss: expose salt len in the verifyingkey

baloo:
Some vendor serialization (NDA :() of signature and public keys will need the size of the salt that was used for signature. Sadly this is only exposed in the signing key (which may be out of reach (HSM)).

### DIFF
--- a/src/pss/verifying_key.rs
+++ b/src/pss/verifying_key.rs
@@ -46,6 +46,11 @@ where
             phantom: Default::default(),
         }
     }
+
+    /// Return specified salt length for this key
+    pub fn salt_len(&self) -> usize {
+        self.salt_len
+    }
 }
 
 //


src/pss/signing_key.rs
```diff
@@ -81,7 +81,7 @@ where
     }
 
     /// Return specified salt length for this key
-    pub fn salt_len(&self) -> usize {
+    pub const fn salt_len(&self) -> usize {
```
tarcieri, line 84: Does it really make sense for this to be `const fn` on a type with no const constructor?

baloo: no it does not. I sent this way too late last night. My bad.

tarcieri (approved):

PR #450, opened 2024-10-07
v0.10.0-rc.0

baloo:


baloo:
I'm bumping cipher, which brings hybrid-array `0.2.0-rc.10` and should bump MSRV.

tarcieri:
I don't think we're ready to call this a `.rc` yet, especially without #394 landed

Issue #418, opened 2024-03-13
Support OAEP decryption with modulus above 4096


GabMus:
Riffing off of #210 I wanted to ask if it's possible to add some feature flag or some other opt-in solution to support keys with modulus size bigger than 4096.

I understand the concerns for algorithmic denial of service, but I think having an escape hatch for when you actually want to use this key size for various reasons would be useful, and in fact it's necessary for a particular need I have :sweat_smile:

tarcieri:
Which type of keys are you asking about? `RsaPublicKey`? PKCS#8 or SPKI encoded keys?

`RsaPublicKey` has [`new_with_max_size`](https://docs.rs/rsa/latest/rsa/struct.RsaPublicKey.html#method.new_with_max_size) and [`new_unchecked`](https://docs.rs/rsa/latest/rsa/struct.RsaPublicKey.html#method.new_unchecked) constructors which can be used to instantiate larger public keys.

PKCS#8 and SPKI are capped to 4096-bits, following [recommendations from the X.509 ecosystem](https://github.com/RustCrypto/RSA/issues/210#issuecomment-1273803217). Since these keys are potentially attacker-controlled they're a potential DoS vector if too large. See #166.

GabMus:
Sorry, I'm in fact referring to PKCS#8 keys in this particular case. Again, as I said I'm well aware of the implication, that's why I was asking specifically for an opt-in solution, something you'd have to explicitly enable. Can this be done?

tarcieri:
You can parse the PKCS#8 key yourself using the `pkcs1` crate and call the appropriate unchecked constructors.

Adding a dedicated feature for this would be rather tricky and seems generally questionable.

GabMus:
In this case it's for keys provided by third parties and we cannot change it.

Regardless the problem does not occur when initializing the `RsaPrivateKey` but when doing the decryption. Let me show you some example code:

```rust
let key = RsaPrivateKey::from_pkcs8_encrypted_pem(my_pem, my_password)?;
let decrypted = key.decrypt(my_oaep_padding_instance, my_encrypted_payload)?; // this errors out with ModulusTooLarge
```

In particular the error seems to come from `RsaPrivateKey::decrypt`, calling `Oaep::decrypt`, calling `key::check_public)`.

This last one then calls `key::check_public_with_max_size`, and that's where we find the size check code:

```rust
    if public_key.n().bits() > max_size {
        return Err(Error::ModulusTooLarge);
    }
```

[...]

PR #453, opened 2024-10-09
feat: add uncheck_modulus_size feature

shuai132:
Same to #418, some projects use a modulus size larger than 4096. And the source code may be written in C++ or Java.
There are still some issues when using the decrypt method, for example, Pkcs1v15Encrypt::decrypt.

Therefore, I have implemented a feature called `uncheck_modulus_size`, hoping it will be useful for others.


tarcieri:
This should be fixed with explicit APIs for large moduli, not a cargo feature that can cause spooky-action-at-a-distance among other dependencies that aren't expecting it

shuai132:
Thanks,
I just tested use openssl(c++), got same result. I will use my fork just fine! thx

PR #455, opened 2024-10-23
feat: updates subtle dep

SonOfMosiah:


tarcieri:
The `subtle` version in Cargo.toml reflects the earliest compatible version.

Cargo will automatically select the latest version.

Issue #445, opened 2024-08-09
Minimum modulus size


ryancdotorg:
RSA is catastrophically weak with 512 bit keys, and anything less than 2048 bits is insecure. A minimum key size should be enforced. Golang has proposed to require 1024 bits minimum:

https://github.com/golang/go/issues/68762

however 2048 would be a good choice where backwards compatibility requirements are minimal

tarcieri:
See also: #350

dignifiedquire:
For testing and supporting existing implementations I don't think a blanket limit makes sense. 
But maybe a default generator that limits the range would be helpful for newcomers.

ryancdotorg:
Not supporting existing implementations which would require this is a benefit.

RSA 1024 and even 2048 are plenty fast for testing.

dignifiedquire:
> Not supporting existing implementations which would require this is a benefit.

I am glad for you that this is how you can run your projects, it is not the same for all of us.

str4d:
I agree with @ryancdotorg and with the rationales outlined in the Go proposal. 

I also note that the `rsa` crate already enforces limits on key sizes, specifically `RsaPublicKey::new` which returns an error for any key larger than 4096 bits. This method has a trapdoor for configurability via `RsaPublicKey::new_with_max_size`.

An intermediate pathway would therefore be to do the same thing for minimum sizes:
- Have the default pathway enforce a secure minimum size in `RsaPrivateKey::new` and `RsaPublicKey::new`, which we can increase as necessary in line with best practices.
- Have a trapdoor for configurability via e.g. `RsaPrivateKey::new_with_insecure_key_size`, which enables reviewers to clearly identify codepaths that are insecure in production.

Then the question of whether to expose the insecure / hazmat configuration APIs at allcan be separated from the "stop people shooting themselves in the foot because they think they are getting 256-bit security" issue.

ryancdotorg:
For reference, here's an example of use of 512 bit RSA that could have damaged a power grid:

https://rya.nc/vpp-hack.html

It's been known to basically worthless in terms of security for decades, let it die.

pinkforest:
The API IMO probably should use mis-use resistance. Just like with x25519-dalek we added static_secrets opt-in to track it's use.

One way would be to split the constructor for insecure and secure constructions - the insecure constructor for given bit-lengths would be hidden behind a feature.

[...]

PR #459, opened 2024-11-13
fix: handle odd keys

dignifiedquire:
Ther are two issues with oddly formed keys that were not properly handled

- avoid using `-` to avoid a subtraction with overflow for pkcs
- always validate the key in `from_components` to avoid errors in the internal `precompute`


tarcieri:
I’m reminded of #445

dignifiedquire:
yes, but the goal here is not general security, but rather for a first step remove unusable keys

once we have a solution for #445 this can be folded in

dignifiedquire:
already published on the `0.9.x` branch, so merging

PR #462, opened 2024-12-01
Fix 2049 bit RSA issues and roundtrip PKCS#1 and PKCS#8

Fethbita:
Fixes issues in tests in https://github.com/RustCrypto/RSA/pull/394

The problem exists because BigUint usually have leading zeroes and the equality checks fail because of that. Now the PKCS#1 and PKCS#8 tests parse the hex number into a BigUint and the equality is checked between them. As for the 2049 bit RSA signing/verification issue, that also was caused by leading zeroes in the left_pad method.


dignifiedquire:
thanks a lot! will take a closer look when back on my computer

dignifiedquire:
merged manually, thank you again!

PR #467, opened 2025-01-30
oaep: support non-string labels

baloo:
This rework oaep to support non-string labels.
One use-case is encryption of secrets in TPM.

https://trustedcomputinggroup.org/wp-content/uploads/TPM-2.0-1.83-Part-1-Architecture.pdf#page=297
Section B.4 RSAES_OAEP
```
For RSA keys protecting a secret value (such as, an encryption key or a session secret), the L parameter
is a byte stream, the last byte of which must be zero, indicating the intended use of the encrypted value. 
```

That would look like:
```rust
    let encrypted_seed = {
        let padding = Oaep::new_with_label::<EkHash, _>(b"IDENTITY\0".to_vec());
        let enc_data = ek_public
            .encrypt(&mut rng, padding, &random_seed[..])
            .expect("failed to encrypt");
        enc_data
    };
 ```

### DIFF
--- a/src/algorithms/oaep.rs
+++ b/src/algorithms/oaep.rs
@@ -1,6 +1,6 @@
 //! Encryption and Decryption using [OAEP padding](https://datatracker.ietf.org/doc/html/rfc8017#section-7.1).
 //!
-use alloc::string::String;
+use alloc::boxed::Box;
 use alloc::vec::Vec;
 
 use digest::{Digest, DynDigest, FixedOutputReset};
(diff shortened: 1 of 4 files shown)


baloo:
I'm pulling this PR in https://github.com/parallaxsecond/rust-tss-esapi/pull/563

tarcieri:
@baloo can you rebase?

tarcieri (approved):

PR #474, opened 2025-02-13
v0.10.0-pre.4

tarcieri:
Notably includes #394 which migrates to `crypto-bigint`


tarcieri:
Note: at some point I think we should downgrade the dependencies and cut a stable release, but in the interim this will make it possible to integration test the new implementation against e.g. the `ssh-*` crates

PR #475, opened 2025-02-13
Use a direct formula to calculate `log(2^x)`

fjarri:
Replaces the current approximate `logf()` which we don't really need for this specific form of the argument.

Note: I am not sure why we calculate `log(2^floor(bit_size / nprimes))` while we could just as well calculate `log(2^(bit_size/nprimes))`, but I left that part intact.


tarcieri (approved):

tarcieri:
Sidebar: @fjarri any thoughts on including some prime counting functionality in `crypto-primes`? I believe there are some more recent optimized algorithms

fjarri:
I think it will be appropriate there. If you have some specific formulas/papers in mind, could you make an issue?

PR #477, opened 2025-02-14
test: implement main test vectors from wycheproof

dignifiedquire:
Thanks to @ctz who implemented a basic structure to run these in in graviola, I ported running against the relevant test vectors from https://github.com/C2SP/wycheproof

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -39,6 +39,8 @@ jobs:
           - stable
     steps:
       - uses: actions/checkout@v4
+        with:
+          submodules: recursive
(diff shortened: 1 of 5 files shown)


tarcieri (approved):
Curious if [`wycheproof2blb`](https://github.com/RustCrypto/utils/tree/master/wycheproof2blb) could be used to do the extraction, but this is fine for now

dignifiedquire:
> Curious if [`wycheproof2blb`](https://github.com/RustCrypto/utils/tree/master/wycheproof2blb) could be used to do the extraction, but this is fine for now

I didn't know this existed, but I have to be honest, the original json format is much easier to reason about and debug.

Issue #466, opened 2025-01-28
`rand_core` v0.9 support


EricZHANG1688:
So with v 0.9.0 for rand and rand_core, will get a error like 'error[E0277]: the trait bound `ThreadRng: CryptoRngCore` is not satisfied".

tarcieri:
@c410-f3r the nice thing about `CryptoRng` is it provides a type-safe interface for ensuring the supplied RNG is actually safe to use for cryptography. A "just make some bytes" API like that can be easily misused with RNGs which are unsuitable for cryptographic use.

In the next set of releases, we're also providing (and pushing people towards) a non-parameterized random generation APIs which avoids compatibility problems between crate versions and explicit RNGs: https://github.com/RustCrypto/traits/pull/1371

PR #478, opened 2025-02-14
bump `rand_core` from `0.6.4` to `0.9.2`; edition 2024; bump MSRV to 1.85

baloo:
This also bumps the dependency of `signature` from `2.3.0-pre.4` to `2.3.0-pre.5`

This pulls:
 - https://github.com/RustCrypto/crypto-bigint/pull/770
 - https://github.com/RustCrypto/traits/pull/1765
 - https://github.com/RustCrypto/traits/pull/1766
 
Fixes https://github.com/RustCrypto/RSA/issues/466


baloo:
I'll need to make an rc release of `pkcs5` and `pkcs8` to fix the `minimum-versions` check

src/unwrap_rng.rs
```diff
@@ -0,0 +1,28 @@
+#![allow(clippy::unwrap_used)]
+use rand_core::{CryptoRng, RngCore, TryCryptoRng, TryRngCore};
+
+/// Wrapper around [`TryRngCore`] implementation which implements [`RngCore`]
+/// by panicking on potential errors.
+///
+/// Similar to [`rand_core::UnwrapErr`] but works with inner references.
+#[derive(Debug, Eq, PartialEq, Hash)]
+pub(crate) struct UnwrapRng<'r, R: TryRngCore>(pub &'r mut R);
```
baloo, line 9: This is only here because [`signature::RandomizedDigestSigner`](https://docs.rs/signature/2.3.0-pre.5/signature/trait.RandomizedDigestSigner.html#tymethod.try_sign_digest_with_rng) forces us to accept a `TryCryptoRng` but the inner rsa implementation for pss and pkcs1.5 only work with `CryptoRng`.

baloo: Yeah, and maybe we should upstream that?

I have to bring it over in [ecdsa too](
https://github.com/RustCrypto/signatures/pull/901/files#diff-a5d3287dfb8311c23907e9e235831d4c651c1e73e69a3153c51b9db1e292f9ef)

baloo: Let's see upstream's opinion on that: https://github.com/rust-random/rand/pull/1589

newpavlov: Maybe it's better to switch the internal implementation to `TryCryptoRng` instead?

baloo: we might need to tweak a bit the [RandomizedPrehashSigner](https://docs.rs/signature/2.3.0-pre.5/signature/hazmat/trait.RandomizedPrehashSigner.html) to accept `TryCryptoRng` then

newpavlov: Sounds good. It should be a simple enough tweak considering hat the methods return `Result`.

baloo: I'm not sure I can change the inner implementation of RSA to accept a `TryCryptoRng`. This calls into `crypto-bigint` and this would require a larger redesign than I'm comfortable with.

src/traits/padding.rs
```diff
@@ -32,7 +32,7 @@ pub trait PaddingScheme {
 /// Digital signature scheme.
 pub trait SignatureScheme {
     /// Sign the given digest.
-    fn sign<Rng: CryptoRngCore>(
+    fn sign<Rng: CryptoRng>(
```
newpavlov, line 35: I think you should use `TryCryptoRng` in all functions/methods which return `Result`.

baloo: But I can't do that because the rng is passed down to `crypto-bigint` which expects a `CryptoRng`. I can `UnwrapMut` on the rng, but that will just throw runtime panics.

I think I'd rather have the consumer do the `UnwrapMut` themselves.

baloo: specifically, we always end up in:
https://github.com/RustCrypto/RSA/blob/9e7faa42bf84ee5a8487012303cf50f297f91f4c/src/algorithms/rsa.rs#L34

Which then calls:
https://github.com/RustCrypto/RSA/blob/9e7faa42bf84ee5a8487012303cf50f297f91f4c/src/algorithms/rsa.rs#L145

which calls into crypto-bigint:
https://github.com/RustCrypto/RSA/blob/9e7faa42bf84ee5a8487012303cf50f297f91f4c/src/algorithms/rsa.rs#L161

And I don't have a `try_random_mod` in crypto-bigint. It specifically wants an `RngCore`

newpavlov: I don't see any obstacles to adding `try_random_mod`. It should be done either way.

baloo: yeah, you're right, it's actually fairly easy.
https://github.com/RustCrypto/crypto-bigint/pull/770

baloo:
I believe this is now ready to merge.

baloo:
nightly doesn't have `edition2024` ?! This is going to be interesting.

newpavlov:
You need to update the pinned nightly version [here](https://github.com/RustCrypto/RSA/blob/master/.github/workflows/ci.yml#L74).

benches/key.rs
```diff
@@ -21,12 +21,20 @@ const DECRYPT_VAL: &str = "\
...
-    let d = hex!("4b97dad7216607064b0d721a431f381e2b6d98524a2095bc1e6bd5ec39c6c9ec3450b2d5db9c328ef3a3d7a11b63eaf57d84f2341159f67e25d917d607427e20a34a41c3c6df8b71e0d9159d85f0ed9bc17345eec140374aef11b2cd638e0c901ee382ff5cfebb3c63290b672fcd1c7ef59ad799b0ed90d49a121ee98587df5cc161c584bc5887ae2a15e787e86ab1e803366150561e0b3b3ae28ebdcf32cd46dff317ed3e1b7590cc300d1d57c9288462d06d9fbe097e52b70dc4fca313ae09906e5fab0c24729b54fe35cc38fe1496419a902f35f08460952bd4783e0e930ba8b520f83eafe6fa6589bbab6e4f4bc5c285672c99f5055eec6a2a30b06e786b");
+    let n = hex!(
+        "7163c842b2190a8970942b2764aed42d4124647b6f30e09a2da1c0e256aa2ee24e790c40c96a4bd66d75c371a915e0703c476b4e1a06f1bd38c5a3c10ae3bd30f4ef62a5aa4f512ad145a06c48e96469a22ce8e621e052f0669a8c34155512d82e55447f0b7e18da94bd911ac7b3aabe706843668964593ee71b2e5e484bcf0c7834101ab5d61bba1e63e6237af40489ce36a260dab70add4fbec24d659db0f7cac099b0a3aa4549acde7fc858a793a975e6cf65ca276b743525f0883980f6ad069bec346d787797386d50fe0c9734be967c7d84ae5b8f349b094079457c0c0c6fee34c42a0b832603804f71e49f3320081637512c6cbf2bb81b6f6be239846d"
```
newpavlov, line 25: It's probably worth to split these strings into several lines.

src/oaep.rs
```diff
@@ -151,7 +151,7 @@ impl PaddingScheme for Oaep {
         )
     }
 
-    fn encrypt<Rng: CryptoRngCore>(
+    fn encrypt<Rng: TryCryptoRng>(
```
newpavlov, line 154: Add `?Sized` bounds like in other methods?

src/key.rs
```diff
@@ -169,7 +169,7 @@ impl PublicKeyParts for RsaPublicKey {
 
 impl RsaPublicKey {
     /// Encrypt the given message.
-    pub fn encrypt<R: CryptoRngCore, P: PaddingScheme>(
+    pub fn encrypt<R: CryptoRng, P: PaddingScheme>(
```
newpavlov, line 172: Same here.

src/oaep/decrypting_key.rs
```diff
@@ -111,15 +111,24 @@ mod tests {
...
         let tokens = [
-            Token::Struct { name: "DecryptingKey", len: 4 },
+            Token::Struct {
+                name: "DecryptingKey",
+                len: 4,
+            },
             Token::Str("inner"),
-            Token::Str("3054020100300d06092a864886f70d01010105000440303e020100020900c9269f2f225eb38d020301000102086ecdc49f528812a1020500d2aaa725020500f46fc249020500887e253902046b4851e1020423806864"),
+            Token::Str(
+                "3054020100300d06092a864886f70d01010105000440303e020100020900c9269f2f225eb38d020301000102086ecdc49f528812a1020500d2aaa725020500f46fc249020500887e253902046b4851e1020423806864",
```
newpavlov, line 120: Split into multiple lines using `\`?

src/pkcs1v15.rs
```diff
@@ -36,7 +36,7 @@ use crate::traits::{PaddingScheme, PublicKeyParts, SignatureScheme};
 pub struct Pkcs1v15Encrypt;
 
 impl PaddingScheme for Pkcs1v15Encrypt {
-    fn decrypt<Rng: CryptoRngCore>(
+    fn decrypt<Rng: TryCryptoRng>(
```
newpavlov, line 39: Add `?Sized`?

src/pkcs1v15/decrypting_key.rs
```diff
@@ -59,17 +59,22 @@ mod tests {
...
         let tokens = [
-            Token::Struct { name: "DecryptingKey", len: 1 },
+            Token::Struct {
+                name: "DecryptingKey",
+                len: 1,
+            },
             Token::Str("inner"),
-            Token::Str("3054020100300d06092a864886f70d01010105000440303e020100020900c9269f2f225eb38d020301000102086ecdc49f528812a1020500d2aaa725020500f46fc249020500887e253902046b4851e1020423806864"),
+            Token::Str(
+                "3054020100300d06092a864886f70d01010105000440303e020100020900c9269f2f225eb38d020301000102086ecdc49f528812a1020500d2aaa725020500f46fc249020500887e253902046b4851e1020423806864",
```
newpavlov, line 76: Split string?

src/pkcs1v15/signing_key.rs
```diff
@@ -49,7 +49,7 @@ where
     }
 
     /// Generate a new signing key with a prefix for the digest `D`.
-    pub fn random<R: CryptoRngCore>(rng: &mut R, bit_size: usize) -> Result<Self> {
+    pub fn random<R: CryptoRng>(rng: &mut R, bit_size: usize) -> Result<Self> {
```
newpavlov, line 52: Add `?Sized`?

src/pkcs1v15/signing_key.rs
```diff
@@ -300,17 +310,17 @@ mod tests {
...
 
         let mut rng = ChaCha8Rng::from_seed([42; 32]);
         let priv_key = crate::RsaPrivateKey::new(&mut rng, 64).expect("failed to generate key");
         let signing_key = SigningKey::<Sha256>::new(priv_key);
 
-        let tokens = [
-            Token::Str("3054020100300d06092a864886f70d01010105000440303e020100020900c9269f2f225eb38d020301000102086ecdc49f528812a1020500d2aaa725020500f46fc249020500887e253902046b4851e1020423806864")
-        ];
+        let tokens = [Token::Str(
+            "3054020100300d06092a864886f70d01010105000440303e020100020900c9269f2f225eb38d020301000102086ecdc49f528812a1020500d2aaa725020500f46fc249020500887e253902046b4851e1020423806864",
```
newpavlov, line 322: Split string?

src/pss.rs
```diff
@@ -86,7 +86,7 @@ impl Pss {
 }
 
 impl SignatureScheme for Pss {
-    fn sign<Rng: CryptoRngCore>(
+    fn sign<Rng: TryCryptoRng>(
```
newpavlov, line 89: Add `?Sized`?

baloo:
tests fixed \o/

src/pss/blinded_signing_key.rs
```diff
@@ -56,13 +56,13 @@ where
     /// Create a new random RSASSA-PSS signing key which produces "blinded"
     /// signatures.
     /// Digest output size is used as a salt length.
-    pub fn random<R: CryptoRngCore>(rng: &mut R, bit_size: usize) -> Result<Self> {
+    pub fn random<R: CryptoRng>(rng: &mut R, bit_size: usize) -> Result<Self> {
```
newpavlov, line 59: Add `?Sized`?

src/pss/blinded_signing_key.rs
```diff
@@ -257,18 +275,18 @@ mod tests {
...
         let mut rng = ChaCha8Rng::from_seed([42; 32]);
         let signing_key = BlindedSigningKey::<Sha256>::new(
             RsaPrivateKey::new(&mut rng, 64).expect("failed to generate key"),
         );
 
-        let tokens = [
-            Token::Str("3054020100300d06092a864886f70d01010105000440303e020100020900c9269f2f225eb38d020301000102086ecdc49f528812a1020500d2aaa725020500f46fc249020500887e253902046b4851e1020423806864")
-        ];
+        let tokens = [Token::Str(
+            "3054020100300d06092a864886f70d01010105000440303e020100020900c9269f2f225eb38d020301000102086ecdc49f528812a1020500d2aaa725020500f46fc249020500887e253902046b4851e1020423806864",
```
newpavlov, line 288: Split string?

src/pss/signing_key.rs
```diff
@@ -63,12 +63,12 @@ where
 
     /// Generate a new random RSASSA-PSS signing key.
     /// Digest output size is used as a salt length.
-    pub fn random<R: CryptoRngCore>(rng: &mut R, bit_size: usize) -> Result<Self> {
+    pub fn random<R: CryptoRng>(rng: &mut R, bit_size: usize) -> Result<Self> {
```
newpavlov, line 66: Add `?Sized`?

src/pss/signing_key.rs
```diff
@@ -282,17 +294,17 @@ mod tests {
...
 
         let mut rng = ChaCha8Rng::from_seed([42; 32]);
         let priv_key = crate::RsaPrivateKey::new(&mut rng, 64).expect("failed to generate key");
         let signing_key = SigningKey::<Sha256>::new(priv_key);
 
-        let tokens = [
-            Token::Str("3054020100300d06092a864886f70d01010105000440303e020100020900c9269f2f225eb38d020301000102086ecdc49f528812a1020500d2aaa725020500f46fc249020500887e253902046b4851e1020423806864")
-        ];
+        let tokens = [Token::Str(
+            "3054020100300d06092a864886f70d01010105000440303e020100020900c9269f2f225eb38d020301000102086ecdc49f528812a1020500d2aaa725020500f46fc249020500887e253902046b4851e1020423806864",
```
newpavlov, line 306: Split string?

src/traits/padding.rs
```diff
@@ -13,15 +13,15 @@ pub trait PaddingScheme {
     ///
     /// If an `rng` is passed, it uses RSA blinding to help mitigate timing
     /// side-channel attacks.
-    fn decrypt<Rng: CryptoRngCore>(
+    fn decrypt<Rng: TryCryptoRng>(
```
newpavlov, line 16: Add `?Sized`?

newpavlov (commented):
The comments below are not relevant to this PR per se, but it could be worth to fix them while we are at it.

src/oaep/decrypting_key.rs
baloo: used `concat!` instead.

newpavlov (approved):

baloo:
huum, wait I also need to pull https://github.com/RustCrypto/traits/pull/1766

Snowiiii:
Seems like everything is done, Or is something missing ?

PR #473, opened 2025-02-11
add comment to specify the rand version

Its-Just-Nans:
Hello,

This PR is not super useful BUT can be useful.

Of course this PR should be kind of reversed/updated when the rand 0.9 will be supported

Linked to 
- https://github.com/RustCrypto/RSA/issues/466
- https://github.com/RustCrypto/traits/issues/1642
- https://rustcrypto.zulipchat.com/#narrow/channel/260047-RSA/topic/rand.20dependencies/near/499002018

Feel free to close if not useful enough to be merged

PR #481, opened 2025-02-25
Document specific version requirement of rand for compiling example code in README.me

hazzery:
The example code provided does not compile using the latest version of rand (0.9). Simply documenting this in the README should reduce some headache in the mean time before #466 has be solved.


tarcieri:
This is more or less a dup of #473

PR #492, opened 2025-03-12
Fix `manual_div_ceil` Clippy lint

newpavlov:


tarcieri:
Hmm, why wasn't clippy catching this in CI?

tarcieri:
Really... huh, I swear I've seen this lint on stable but maybe I'm mistaken

newpavlov:
Hm, it was [added](https://rust-lang.github.io/rust-clippy/master/index.html#/manual_div_ceil) in 1.83, but maybe not enabled by default yet? Running `cargo +stable clippy` does not trigger the warnings.

tarcieri:
Aah, perhaps it's being enabled by other lints like `pedantic`

PR #498, opened 2025-03-12
Make SigningKey::from use Self::new()

marshallpierce:
This matches https://github.com/RustCrypto/RSA/pull/472.


tests/pkcs1v15.rs
```diff
@@ -21,3 +25,55 @@ fn signature_stringify() {
     assert_eq!(format!("{:X}", signature), expected);
     assert_eq!(signature.to_string(), expected);
 }
+
+#[test]
```
tarcieri, line 29: To use PEM here you'll have to gate this on the `pem` feature as above:

```suggestion
#[cfg(feature = "pem")]
#[test]
```

PR #499, opened 2025-03-12
use a public url for git submodules

baloo:
When pulling `RSA` with a `[patch.crates-io]`, the CI runner doesn't carry an SSH key with github and fails to clone the submodules.

This changes the `git@github.com` to an `https://github.com` to allow anonymous clones.


baloo:
Example: https://github.com/RustCrypto/formats/actions/runs/13823344798/job/38673496298?pr=1689#step:5:23

tarcieri (approved):

PR #504, opened 2025-04-13
Bump pin for crypto-primes

baloo:


tarcieri (approved):

tarcieri:
Hmm, weird, the tests legitimately seem to be hanging:

```
test algorithms::generate::tests::test_impossible_keys has been running for over 60 seconds
```

cc @fjarri @dvdplm

fjarri:
It looks like it replaced finding a regular prime with finding a safe
prime, which is much slower

PR #506, opened 2025-04-23
Shorten widths of `BoxedUint`s

### LINKED ISSUE #490: Slowdown after switch to `crypto-bigint`
Benchmarks slowed down several times after #394. What could be the reason?

Performance improvements for boxed uints (https://github.com/RustCrypto/crypto-bigint/pull/777) and the corresponding changes in `crypto-primes` (https://github.com/entropyxyz/crypto-primes/pull/78) don't really change the results much according to my tests.

`sign` test time is split ~50/50 between Montgomery exponentiation (with all the time spent in `almost_montgomery_mul()`, the lowest level function) and `BoxedUint::inv_mod()`; `decrypt` is almost exclusively exponentiation. Both of these are the calls in `RSA` itself; `crypto-primes` calls take negligible time. 

So it seems that either these two functions are somehow much slower than `num-bigint` (possible, yet unlikely, and straightforward to test), or somehow #394 changed the algorithm to apply them more times than necessary.

fjarri:
Fixes #490 

- Bump `crypto-bigint` dependency to the current trunk.
- Use the width of a single prime instead of the full modulus where possible. Brings `bench_rsa_2048_pkcsv1_decrypt` to pre-`crypto-bigint` values.
- Add a check to `RsaPrivateKey::from_components()` to ensure consistency between the primes and the modulus.
- Remove zero padding limbs from the primes and the modulus in `RsaPrivateKey::from_components()`.
- Add a check to `rsa_decrypt()` to ensure the bit precision of the ciphertext is the same as that of the modulus.

Notes:
- `bench_rsa_2048_pkcsv1_sign_blinded` can be restored to the original performance by using variable-time inversion in `algorithms::rsa::blind()` (as it was during `num-bigint` times), but it seems to me that the blinding factor must be kept secret, so we have to use the constant-time inversion. This leads to about 5x slowdown compared to pre-`crypto-bigint` performance.  
- The changes in test vectors are due to https://github.com/RustCrypto/crypto-bigint/pull/781

### Possible further improvements
- Keep precomputed values as `Odd`/`NonZero` as appropriate.
- If https://github.com/RustCrypto/crypto-bigint/issues/811 is fixed, resizing integers for the purposes of division can be avoided.


tarcieri:
> The r^(-1) mod n was variable-time before crypto-bigint; is that secure?

Everything was variable-time before `crypto-bigint`, and to my knowledge it's insecure. It would probably be best to make it constant time despite the performance hit.

Hopefully we'll get better performance after a migration to bingcd

tarcieri:
> - The new try_set_precision() may be a method in BoxedUint
> - shorten/widen may need some fallible equivalents

So if I understand the goal of these, it's to prevent truncating the contents of a value, correct?

If so, adding them upstream seems fine. Perhaps they could be made constant-time?

fjarri:
> Everything was variable-time before crypto-bigint, and to my knowledge it's insecure. It would probably be best to make it constant time despite the performance hit.

That's unfortunate, it's like a 5x slowdown. 

> So if I understand the goal of these, it's to prevent truncating the contents of a value, correct?

Yep.

tarcieri:
@fjarri do you want to try to upstream `try_set_precision` and e.g. `try_shorten`/`try_widen` first, or merge this as-is and follow up on that later?

fjarri:
Let me try it out, I'll see if I can distill some required functionality into general-use methods.

As for the inversion, I need to check if preparing the inverter in advance makes things any better.

tarcieri:
It’s unlikely to impact performance aside from optimizing away a single allocation

fjarri:
Converting to draft until we merge https://github.com/RustCrypto/crypto-bigint/pull/809 in some form

src/key.rs
```diff
@@ -789,14 +822,14 @@ mod tests {
...
+            "3056020100300d06092a864886f70d010101050004423040020100020900a",
+            "b240c3361d02e370203010001020811e54a15259d22f9020500ceff5cf302",
+            "0500d3a7aaad020500ccaddf17020500cb529d3d020500bb526d6f"
         ))];
         assert_tokens(&priv_key.clone().readable(), &priv_tokens);
 
         let priv_tokens = [Token::Str(
-            "3024300d06092a864886f70d01010105000313003010020900aecdb5fae1b092570203010001",
+            "3024300d06092a864886f70d01010105000313003010020900ab240c3361d02e370203010001",
         )];
```
tarcieri, line 833: What are these changes? (and many similar ones in the other key files)

fjarri: Caused by the `crypto-bigint` bump, namely https://github.com/RustCrypto/crypto-bigint/pull/781

tarcieri (approved):

PR #510, opened 2025-04-30
Bump to signature 3

baloo:
alternative to https://github.com/RustCrypto/RSA/pull/505

### DIFF
--- a/src/pkcs1v15.rs
+++ b/src/pkcs1v15.rs
@@ -458,15 +458,15 @@ mod tests {
         let signing_key = SigningKey::<Sha1>::new(priv_key);
 
         for (text, expected) in &tests {
-            let out = signing_key.sign(text.as_bytes()).to_bytes();
+            let sig: Signature = signing_key.sign(text.as_bytes());
+            let out = sig.to_bytes();
             assert_ne!(out.as_ref(), text.as_bytes());
             assert_ne!(out.as_ref(), &Sha1::digest(text.as_bytes()).to_vec());
             assert_eq!(out.as_ref(), expected);
 
             let mut rng = ChaCha8Rng::from_seed([42; 32]);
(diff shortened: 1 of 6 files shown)


src/pkcs1v15.rs
```diff
@@ -458,15 +458,15 @@ mod tests {
         let signing_key = SigningKey::<Sha1>::new(priv_key);
 
         for (text, expected) in &tests {
-            let out = signing_key.sign(text.as_bytes()).to_bytes();
+            let sig: Signature = signing_key.sign(text.as_bytes());
```
tarcieri, line 461: Is this really still needed?

baloo:
```
error[E0282]: type annotations needed
   --> src/pkcs1v15.rs:461:35
    |
461 |             let out = signing_key.sign(text.as_bytes()).to_bytes();
    |                                   ^^^^
    |
help: try using a fully qualified path to specify the expected types
    |
461 -             let out = signing_key.sign(text.as_bytes()).to_bytes();
461 +             let out = <pkcs1v15::signing_key::SigningKey<CoreWrapper<Sha1Core>> as Signer<S>>::sign(&signing_key, text.as_bytes()).to_bytes();
    |

For more information about this error, try `rustc --explain E0282`.
```

Without it.

In the end I think #505 creates less API breakage.

tarcieri: I think the blanket impl is legitimately breaking inference for the case there's an explicit `Signer` impl instead of using the blanket one, which would be a case of https://github.com/RustCrypto/traits/issues/1831

baloo: agreed

baloo:
Closed in favor of https://github.com/RustCrypto/RSA/pull/513

PR #519, opened 2025-05-22
PublicKeyParts: provide serialization helpers

### LINKED ISSUE #518: usability: exponent serialization
When serializing exponent, and since we moved to `crypto-bigint`, `BoxedUint::to_be_bytes()` now prepend the bytes with zeroes.

This causes issues when interacting with other crypto libraries like ring which do not expect zero prepended to the exponent.

I've made a reproducer here:
 https://github.com/baloo/rsa-exponent-serialization

`main` is using `rsa 0.9`

Here are the changes I assumed I would need to do when upgrading to 0.10:
https://github.com/baloo/rsa-exponent-serialization/compare/baloo/rsa0.10.broken

Here what I ended up having to do:
https://github.com/baloo/rsa-exponent-serialization/compare/baloo/rsa0.10

baloo:
When serializing the exponent or the modulus from a key, `BoxedUint::to_be_bytes` will return a slice padded with zeroes on the left because of its inner representation. Not every implementation out there will handle leading zeroes.

This provides a `PublicKeyParts::n_bytes`/`e_bytes` that will strip out those leading zeros.

See https://github.com/RustCrypto/RSA/issues/518 for more context.

Depends:
  - https://github.com/RustCrypto/crypto-bigint/pull/824

Fixes #518

### DIFF
--- a/src/traits/keys.rs
+++ b/src/traits/keys.rs
@@ -1,5 +1,6 @@
 //! Traits related to the key components
 
+use alloc::boxed::Box;
 use crypto_bigint::{
     modular::{BoxedMontyForm, BoxedMontyParams},
     BoxedUint, NonZero,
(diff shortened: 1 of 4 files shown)


src/traits/keys.rs
```diff
@@ -27,6 +28,22 @@ pub trait PublicKeyParts {
     fn n_bits_precision(&self) -> u32 {
         self.n().bits_precision()
     }
+
+    /// Returns the big endian serialization of the modulus of the key
+    fn n_bytes(&self) -> Box<[u8]> {
+        let modulus = self.n();
+        let modulus_zeros = modulus.leading_zeros() as usize / 8;
+        let modulus = modulus.to_be_bytes();
+        (&modulus[modulus_zeros..]).into()
```
tarcieri, line 37: This is the logic it would be good to extract into `BoxedUint` (possibly e.g. `to_be_bytes_trimmed`?)

tarcieri (approved):

tarcieri (approved):

PR #529, opened 2025-06-14
Fix modulus size check in `check_public_with_max_size`

### LINKED ISSUE #528: `RsaPublicKey::new_with_max_size` behaves differently
On `rsa 0.9.8`, `RsaPublicKey::new_with_max_size` treated the passed `max_size` as the actual maximum number of bits the modulus is allowed to have; however `rsa 0.10.0-rc.0` expects that same parameter to actually be the maximum number of "bits precision" the modulus is allowed to have.

It would be nice if the documentation were at least amended to make this clear.

**`rsa 0.9.8`**:
```rust
use rsa::{BigUint, RsaPublicKey};
fn main() {
    let n = BigUint::from_bytes_be([1, 116, 2, 177, 35, 163, 109, 146, 185].as_slice());
    assert!(RsaPublicKey::new_with_max_size(n, 3u32.into(), 65).is_ok());
}
```

**`rsa 0.10.0-rc.0`**:
```rust
use rsa::{BoxedUint, RsaPublicKey};
fn main() {
    let n =
        BoxedUint::from_be_slice([1, 116, 2, 177, 35, 163, 109, 146, 185].as_slice(), 65).unwrap();
    // Despite `n` being 65 bits, `max_size` is considered to be too small.
    assert!(RsaPublicKey::new_with_max_size(n.clone(), 3u32.into(), 65).is_err());
    // `128` is the smallest `max_size` that doesn't cause an error which happens to also be
    // the value of `n.bits_precision()`.
    assert!(RsaPublicKey::new_with_max_size(n, 3u32.into(), 128).is_ok());
}
```

```bash
[zack@laptop ~]$ rustc --version
rustc 1.87.0 (17067e9ac 2025-05-09)
[zack@laptop ~]$ uname -a
Linux laptop 6.15.1-arch1-2 #1 SMP PREEMPT_DYNAMIC Sat, 07 Jun 2025 14:36:24 +0000 x86_64 GNU/Linux
```

tarcieri:
It was checking the bits of precision, rather than the actual bit length of the key.

Closes #528.

### DIFF
--- a/src/key.rs
+++ b/src/key.rs
@@ -622,7 +622,7 @@ pub fn check_public(public_key: &impl PublicKeyParts) -> Result<()> {
 /// Check that the public key is well formed and has an exponent within acceptable bounds.
 #[inline]
 fn check_public_with_max_size(n: &BoxedUint, e: &BoxedUint, max_size: usize) -> Result<()> {
-    if n.bits() as usize > max_size {
+    if n.bits_vartime() as usize > max_size {
         return Err(Error::ModulusTooLarge);
     }
 
--- a/src/key.rs
+++ b/src/key.rs
@@ -622,7 +622,7 @@ pub fn check_public(public_key: &impl PublicKeyParts) -> Result<()> {
 /// Check that the public key is well formed and has an exponent within acceptable bounds.
 #[inline]
 fn check_public_with_max_size(n: &BoxedUint, e: &BoxedUint, max_size: usize) -> Result<()> {
-    if n.bits_precision() as usize > max_size {
+    if n.bits() as usize > max_size {
         return Err(Error::ModulusTooLarge);
     }
 


src/key.rs
```diff
@@ -622,7 +622,7 @@ pub fn check_public(public_key: &impl PublicKeyParts) -> Result<()> {
 /// Check that the public key is well formed and has an exponent within acceptable bounds.
 #[inline]
 fn check_public_with_max_size(n: &BoxedUint, e: &BoxedUint, max_size: usize) -> Result<()> {
-    if n.bits_precision() as usize > max_size {
+    if n.bits() as usize > max_size {
```
tarcieri, line 625: I suppose it can be this:
```suggestion
    if n.bits_vartime() as usize > max_size {
```

PR #537, opened 2025-07-09
Gate PKCS#1/PKCS#8 support on `encoding` feature

tarcieri:
Support for these formats has a number of dependencies required, and isn't always necessary, e.g. the `ssh-key` crate implements an entirely different encoding format for RSA keys.

Making support an optional feature avoids incurring these dependencies when that's the case.

This commit also removes the `pem` feature and makes PEM support on-by-default whenever `encoding` is enabled. This only adds the `pem-rfc7468` dependency. Most users will probably want PEM support if they have the `encoding` feature enabled.


tarcieri:
Going to merge this so I can work on https://github.com/RustCrypto/formats/pull/1927

PR #561, opened 2025-08-22
Make Tony's tests green again

pinkforest:


tarcieri:
I can rebase my branch... this is a lot

tarcieri:
Rebased

pinkforest:
Re-merged in

Testing takes ~14 mins in github CI with 2048 bits keysize instead of 4 mins 64 bits

It's mainly that one integration test one but units don't take that long.

PR #576, opened 2025-09-09
Add minimum `bit_size` of 1024 to `RsaPrivateKey::new`

tarcieri:
Also adds a `hazmat`-gated `RsaPrivateKey::new_unchecked` as an escape hatch.

Partial step towards addressing #445.

cc @pinkforest


tarcieri:
#526 was probably trying to do too much at once.

This is at least a "stop-the-bleeding" fix that will hopefully deter people from generating insecurely sized keys, at least so easily.

PR #585, opened 2025-09-26
wycheproof: add `git submodule update --init` to error message

tarcieri:
It's otherwise unclear what needs to be done to address these errors in the event the wycheproof submodule has not been initialized


dignifiedquire (approved):
great idea

PR #594, opened 2025-10-15
Support for cryptographic operations with larger keys

hko-s:
Currently, this crate allows instantiation of public keys larger than 4096 bit (via `RsaPublicKey::new_with_max_size`), but doing cryptographic operations with such public keys fails in `key::check_public`, which always checks the modulus size against the constant `RsaPublicKey::MAX_SIZE`.

Also, there's no public API to instantiate larger private keys, or to do cryptographic operations with them.

I think it would be nice to cap both public and private key sizes to 4096 bit by default, but to allow opt-in creation of larger keys (complete with working cryptographic operations).

### DIFF
--- a/src/key.rs
+++ b/src/key.rs
@@ -32,9 +32,6 @@ pub struct RsaPublicKey {
     ///
     /// Typically 0x10001 (65537)
     e: BigUint,
-
-    /// The size limit given at creation time
-    pub(crate) max_size: usize,
 }
 
 /// Represents a whole RSA key, public and private parts.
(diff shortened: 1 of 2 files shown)


hko-s:
> You don't need a `max_size` carrier on the public key. It's checked at construction time, so once constructed there's no need to use it again.

Thanks, sounds good!

I was hesitant to drop the modulus size check that currently happens in `check_public_with_max_size`:

```
    if public_key.n().bits() > max_size {
        return Err(Error::ModulusTooLarge);
    }
```

If we can safely drop that bit of logic when doing cryptographic operations, I'm all for it (but I myself couldn't judge if it is safe).

hko-s:
Awesome, thanks!

Issue #626, opened 2026-01-07
Padding implementation is not constant-time


tarcieri:
This is a followup to #19, which was originally about the modular exponentiation implementation not being constant-time, and also became our general tracking issue for the subsequent [Marvin Attack](https://people.redhat.com/~hkario/marvin/).

We've gone to great lengths in `crypto-bigint` to produce the closest thing we can to a constant-time implementation of modular exponentiation, and while we still need to e.g. verify that's truly the case via static analysis tooling, based on the [latest analysis from the Marvin Toolkit](https://github.com/RustCrypto/RSA/issues/19#issuecomment-3719177009) it seems like the remaining sidechannels in our implementation are probably no longer coming from `crypto-bigint`, but are instead in this crate's implementation of RSA padding modes.

PKCS#1 v1.5 in particular notably has a long history of sidechannels going back to Bleichenbacher's original 1998 attack, and attacks like Marvin can be seen as an evolution of that attack.

This I-D contains guidance for implementing RSA in constant-time, including things like handling depadding errors using strategies like implicit rejection:

https://datatracker.ietf.org/doc/draft-irtf-cfrg-rsa-guidance/

Ryujiyasu:
## Formal Verification of the Timing Vulnerability using Kani

I've built [ne](https://gitlab.com/Ryujiyasu/ne), a formal verification tool that uses [Kani](https://github.com/model-checking/kani) (bounded model checking) to mathematically prove properties of cryptographic code.

I applied ne to the vulnerable `pkcs1v15_encrypt_unpad` pattern from this crate and can confirm:

### Vulnerability Detection

Kani formally proves the timing leak exists — the function has a **secret-dependent branch** (`if valid == 0 { return Err(...) }`) that causes different control flow paths depending on padding validity:

```
Harness: detect_vulnerable_timing_leak
Result:  VERIFICATION:- FAILED
Reason:  "Timing leak: function returns None for invalid padding"
```

This is not a test failure — it's an exhaustive mathematical proof over **all possible inputs** that the function violates constant-time properties.

### Fix Verification (Implicit Rejection)

I implemented a constant-time fix using implicit rejection (same approach as Go's `crypto/rsa` and the [IRTF guidance](https://datatracker.ietf.org/doc/draft-irtf-cfrg-rsa-guidance/)) and Kani verifies:

| Harness | Result | What it proves |
|---------|--------|----------------|
[...]

PR #680, opened 2026-04-01
Implicit rejection api for PKCS#1 v1.5 decryption

lubux:
This draft PR replaces the PKCS#1 v1.5 unpad function with an implicit rejection API based on:

- OpenSSL https://github.com/openssl/openssl/commit/7fc67e0a33102aa47bbaa56533eeecb98c0450f7
- https://datatracker.ietf.org/doc/draft-irtf-cfrg-rsa-guidance/

Issue #626

The new PKCS#1 v1.5 unpad function does not return an error if the padding is invalid. 
Instead, it deterministically generates and returns a replacement random message derived from the secret key and the  ciphertext.
As a result, callers cannot distinguish between valid and invalid padding based on the output, thus reducing the risk side-channel attacks.

Breaking changes:

- Removes the sha2 feature flag because SHA-2 is required for the implicit rejection algorithm.

- PKCS#1 v1.5 decryption no longer returns an error if the padding is invalid.

WARNING: This PR requires careful review before progressing beyond the draft stage.

 ~~TODO: The existing 0 padding function does not seem to be constant time: uint_to_zeroizing_be_pad~~


tarcieri:
To reiterate my [comment from #626](https://github.com/RustCrypto/RSA/issues/626#issuecomment-4199137969):

Instead of adding new APIs, features, etc, I would like to see implicit rejection added by removing the existing fallible handling of padding. If there are APIs that only handle padding errors the `Result` can be changed to return a value, but APIs that are part of existing traits, or APIs that handle errors other than padding should continue to return a `Result`.

Ideally it should be a change that's largely invisible, but we should copiously document that PKCS#1 v1.5 padding errors no longer return an error but instead the pseudorandom rejection symbol.

lubux:
> To reiterate my [comment from #626](https://github.com/RustCrypto/RSA/issues/626#issuecomment-4199137969):
> 
> Instead of adding new APIs, features, etc, I would like to see implicit rejection added by removing the existing fallible handling of padding. If there are APIs that only handle padding errors the `Result` can be changed to return a value, but APIs that are part of existing traits, or APIs that handle errors other than padding should continue to return a `Result`.
> 
> Ideally it should be a change that's largely invisible, but we should copiously document that PKCS#1 v1.5 padding errors no longer return an error but instead the pseudorandom rejection symbol.

Removed the feature flag and made it the default behavior without introducing new APIs.

